PR #46, opened 2023-01-31
Remove comment to fix errors with Whitenoise

hadecolliday:
Whitenoise throws the following error with the comment that seems to hold no value.

```
The JS file 'unfold\forms\js\trix.js' references a file which could not be found:
  unfold/forms/js/trix.umd.min.js.map

Please check the URL references in this JS file, particularly any
relative paths which might be pointing to the wrong location.
```


lukasvinclav:
Hey, thank you very much for the PR. I wanted to merge it but I accidentally solved it in another commit. Anyway, issue is fixed in the main branch and I will make release soon.

PR #50, opened 2023-02-27
feat: Add dynamic sidebar permissions customisation

silox:
This PR introduces sidebar customisation via permissions.

It is now possible to provide visibility callback to "SIDEBAR"."default_visibility" in UNFOLD settings to apply permissions to all sidebar items. You can override visibility of specific items to users by filling "visibility_callback" for each item.

If not item is visible for user, its group title also won't be shown. If "default_visibility_callback" is not specified, it will act as `lambda request: True` (anyone can see everything).

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,13 +182,20 @@ UNFOLD = {
         {
             "models": [
                 "app_label.model_name_in_lowercase",
+                "app_label.model_name2_in_lowercase",
             ],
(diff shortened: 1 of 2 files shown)


mredyk:
I added new logic for tab customization by request. It intuitively works the same as the sidebar permission logic.

PR #51, opened 2023-03-06
fix: add link color

bek9:
This change fixes the color of the add link in the navigation bar so that it is visible. There was an issue with the add link in the navigation bar being white and therefore invisible in light mode. I think it's been missing since #26 or #43.

Before:
Screen Shot 2023-03-06 at 12 01 18
Screen Shot 2023-03-06 at 12 01 23

After:
Screen Shot 2023-03-06 at 12 01 06
Screen Shot 2023-03-06 at 12 01 15

### DIFF
--- a/src/unfold/templates/admin/change_list_object_tools.html
+++ b/src/unfold/templates/admin/change_list_object_tools.html
@@ -7,8 +7,8 @@
 
             <span class="block bg-gray-200 h-5 mr-4 w-px dark:bg-gray-700"></span>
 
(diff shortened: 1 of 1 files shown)


bek9:
I discovered that the bg-primary color doesn't work. I was using an old version of the settings, which says to use hex color codes for `COLORS`, but I noticed that it has changed to RGB values in README.md. I checked and found [this change](https://github.com/remastr/django-unfold/commit/4414a0b2b0daea09861663ebec34d4ace9de1284?short_path=b335630) and I changed the settings to RGB values and all the buttons now display the background color correctly.

PR #80, opened 2023-06-03
feat: test flow

lukasvinclav:


.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  tests:
+    name: Python ${{ matrix.python-version }}
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - 3.8
+          - 3.9
```
PavelPancocha, line 22: @lukasvinclav What about having them all as string?

PR #83, opened 2023-06-07
fix: changelist-filter layout bug

TheVaan:
Hi,

there is a small layout bug: when opening filter in a changelist view, the backdrop has an offset from the right border of the browsers window.

image

Fix:
* remove mr-1

Chore:
* replace {top|bottom|right|left}-0 by inset-0 (same effect with less code)

### DIFF
--- a/src/unfold/templates/admin/change_list.html
+++ b/src/unfold/templates/admin/change_list.html
@@ -117,7 +117,7 @@
                 {% if cl.has_filters %}
                     <input type="checkbox" id="show-filters" class="hidden peer" />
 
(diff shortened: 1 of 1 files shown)


lukasvinclav:
Hi, thanks for update. Tested it and it seems this is will fix the issue but it is needed to compile CSS classes `npm run tailwind:build`

lukasvinclav:
Thank you very much for the fix, but I already pushed a desired change with compiled CSS.

TheVaan:
You're welcome. Sorry for not responding in time - I didn’t get notified about your comment so I missed updating my PR.
Thank you for fixing it!

PR #87, opened 2023-06-21
feat: compat with python3.7

Aleksi44:
Hello,

I'm the author of [django-admin-tailwind](https://github.com/Aleksi44/django-admin-tailwind) (a django admin theme with TailwindCSS and AlpineJS too). 
Congratulations for this great work (much better than my theme). 
I'd like to add compatibility with python3.7 and I'm ready to make the effort to make it happen if it's okay with you.


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 isolated_build = true
 envlist =
-    python{38,39,310}-django{32,41,42}
+    python{37,38,39,310}-django{32,32,41,42}
```
lukasvinclav, line 4: Two times 3.2?

lukasvinclav (commented):
in pyproject.toml we have to change target-version as well

lukasvinclav:
Hey Aleksi, thanks for kind words, always nice to meet fellow Django developer :-P I left few comments in your PR but then I checked for how long 3.7 is going to be officially supported and after quick look I noticed that support is planned for another 6 days :-D I'm not sure if it makes sense to put an energy into this topic. What do you think?

https://endoflife.date/python

Aleksi44:
Hey @lukasvinclav, thanks for your feedback. I've just fixed your comments. 

I work for a company that can't upgrade to python3.8 right now. They want an advanced Django Admin (django-unfold is perfect for what they want). I'd like to contribute directly to this repository rather than maintain a fork if you agree.

I could also bring in some new features later :)

lukasvinclav:
Just tested python 3.7 support and it seems that Poetry requires at least 3.8.

pyproject.toml
```diff
@@ -61,8 +62,9 @@ addopts = """\
 django_find_project = false
 
 [tool.poetry.dependencies]
-python = ">=3.8"
+python = ">=3.7"
 django = ">=3.2"
+importlib-metadata = "6.7.0"
```
Aleksi44, line 67: We need this library for Python3.7 to avoid error :
```
ModuleNotFoundError: No module named 'importlib_metadata'
```
Triggered here : https://github.com/pytest-dev/pluggy/blob/main/src/pluggy/_manager.py#L31C12-L31C30

I don't know if there's a way to install it only for Python3.7?

src/unfold/settings.py
```diff
@@ -39,7 +39,7 @@
 }
 
 
-@lru_cache
+@lru_cache(maxsize=None)
```
Aleksi44, line 42: We need to declare `maxsize=None` for Python3.7 to avoid this error :
```
src/unfold/settings.py:43: in <module>
    def get_config(settings_name=None):
/usr/lib/python3.7/functools.py:490: in lru_cache
    raise TypeError('Expected maxsize to be an integer or None')
TypeError: Expected maxsize to be an integer or None
```

poetry.lock
Aleksi44, line 1: I regenerated the poetry.lock because we need for Python3.7 :
```
importlib-metadata = "6.7.0"
```
I don't know if it's possible to have several poetry.locks in the same project (one specific to Python3.7 and and an other).

Aleksi44 (commented):
details of my changes

lukasvinclav:
@Aleksi44 Regarding the lru_cache, I did some investigation and maybe I would prefer to use it with explicitly setting default values. You can read more here:

https://stackoverflow.com/questions/47218313/use-functools-lru-cache-without-specifying-maxsize-parameter

I think this is the last thing which we are missing and then we are good to go.

Aleksi44:
Hey @lukasvinclav 

I see that the commit you merged is no longer on main. Did you have any problems?

lukasvinclav:
@Aleksi44 I sent you an email last Friday. I used address from your profile.

PR #88, opened 2023-06-26
fix: poetry error when used as a local dependency

minholi:
Tried to use it as described on README.md https://github.com/unfoldadmin/django-unfold#unfold-development but was getting an error.

Using poetry 1.5.1


lukasvinclav:
What error you were getting? Can you post the message here? Thanks

PR #90, opened 2023-06-26
feat: Support django-reversion and django-reversion-compare

justinqian42:
Initial support for both django-reversion and django-reversion-compare plugins, to correctly style the corresponding pages


justinqian42:
I'm trying to apply the same logic for the import-export support without touching any core codes of unfold.
To use this feature, users will need to inherit VersionAdmin and/or CompareVersionAdmin from unfold.contrib.reversion and/or unfold.contrib.reversion_compare.

The corresponding change to the manual will be submitted later after this PR is properly reviewed.

src/unfold/contrib/reversion/admin.py
```diff
@@ -0,0 +1,22 @@
...
+    recover_list_template = "reversion\recover_list.html"
+    recover_form_template = "reversion\recover_form.html"
+    revision_form_template = "reversion\revision_form.html"
+    object_history_template = "reversion\object_history.html"
+
+    actions_list = ["changelist_global_action_reversion_recover"]
+
+    @action(description=_("Recover"), url_path="recover/")
+    def changelist_global_action_reversion_recover(self, request: HttpRequest):
+        # return redirect(self.get_urls)
```
PavelPancocha, line 21: Unnecessary comment

src/unfold/contrib/reversion_compare/admin.py
```diff
@@ -0,0 +1,24 @@
...
+    revision_form_template = "reversion\revision_form.html"
+    object_history_template = "reversion-compare\object_history.html"
+    compare_template = "reversion-compare\compare.html"
+    compare_raw_template = "reversion-compare\compare_raw.html"
+
+    actions_list = ["changelist_global_action_reversion_recover"]
+
+    @action(description=_("Recover"), url_path="recover/")
+    def changelist_global_action_reversion_recover(self, request: HttpRequest):
+        # return redirect(self.get_urls)
```
PavelPancocha, line 23: Unnecessary comment

PavelPancocha (approved):
In overall great! Thank you!

Having some tests would be great, maybe do them as another tests suite with conditinal dependecis for tox?

src/unfold/contrib/reversion/admin.py
```diff
@@ -0,0 +1,22 @@
...
+from django.urls import reverse_lazy
+from django.utils.translation import gettext_lazy as _
+from django.http import HttpRequest
+from unfold.decorators import action
+from reversion.admin import VersionAdmin as BaseVersionAdmin
+
+
+class VersionAdmin(BaseVersionAdmin):
+    change_list_template = "admin\change_list.html"
+    recover_list_template = "reversion\recover_list.html"
```
PavelPancocha, line 12: The back-slashes are wrong. Doesn't work on unix like systems!

PavelPancocha (changes requested):
@justinqian42 

I did deeper review and I found it doesn't work as expected:

- the path with `\` _backslash_ is not valid
_(I fixed that on my side for further testing)_
- Unfold media (CSS) are missing, Can be fixed by adding Unfold "ModelAdmin" as base class (I suppose, that usage should be only with "version" admin.

image

image


Integrations tests would be great (that would test, that all media files etc. are loaded properly). Also that given pages could be displayed, as because of the broken path it didn't work properly (Django crashed).

justinqian42:
Thank you @PavelPancocha for the feedback.
It's been a while since I submitted the PR and I almost forgot it.

To be honest, I've got very little experience with the frontend stuff and class method overwriting. Will try to fix the issues but please don't put a high expectation on me......

Will find some time.

PavelPancocha:
@justinqian42 Feel free to let me know - we look forward to using Django Unfold with reversion in ~2 months, so it would be great to make this working :)

alex-deltia:
@lukasvinclav does the closing of this PR (and the related issue) with the "wontfix" tag mean that you won't be merging any PRs unless it's through your consulting services?

PR #122, opened 2023-09-19
feat: permission check for tabs and navigation

mcosti:
Adding support for permission checks.

Note: This is not tested, it's coded in blind


lukasvinclav:
Hey, looks very good! I will do some testing to see if it works properly and  I will get back to you soon.

PR #184, opened 2023-11-07
Improve admin.py example for register models

Mte90:
As per title, the `register` method as second parameter accept the class so the code can be simplified.


lukasvinclav:
Thanks for the PR. I was thinking about it, but I have some doubts regarding the security. With the code in this PR, it is possible to register models into admin with private information.

Mte90:
It makes sense but we can improve the readme content to explain it.

PR #255, opened 2024-01-12
Enhance Action Lists Based on Target Object for "Detail" and "Submit Line" Actions

dfg-98:
## Summary:
This PR introduces the ability to modify the action lists based on the attributes of the target object or instance for 'Detail' and 'Submit Line' actions. This is achieved by passing the `object_id` parameter to the `get_actions_detail` and `get_actions_submit_line` methods, which are called on `changeform_view`. This allows for custom action filtering based on object attributes.

## Changes:
- The `get_actions_submit_line` and `get_actions_detail` methods now accepts an `object_id` parameter.
- Actions can be added or removed based on the attributes of the object.

## Example:
This example demonstrates how to use the new feature. The submit_save_and_publish action is only included in the action list if the book is not published.

```python
# Inside Model Admin
actions_submit_line = ["submit_save_and_publish"]

@action(description=_("Save & Refer"), permissions=["change"])
    def submit_save_and_publish(self, request: HttpRequest, obj: Book):
        obj.published = True
        obj.save()
        return redirect(
            reverse_lazy(
                f"admin:{obj._meta.app_label}_{obj._meta.model_name}_changelist"
            )
        )


def get_actions_submit_line(self, request, object_id=None):
        actions = super().get_actions_submit_line(request=request, object_id=object_id)

        if not object_id:
            return actions

        book = self.get_object(request, object_id=object_id)
        if not book.published:
            actions = [
                action
                for action in actions
                # Action name based on UnfoldAction
                if action.action_name = f"{self.model._meta.app_label}_{self.model._meta.model_name}_submit_save_and_publish"
            ]

        return actions
```

### DIFF
--- a/src/unfold/admin.py
+++ b/src/unfold/admin.py
@@ -423,7 +423,7 @@ class ModelAdmin(ModelAdminMixin, BaseModelAdmin):
         """
         return [self.get_unfold_action(action) for action in self.actions_list or []]
 
-    def get_actions_detail(self, request: HttpRequest) -> List[UnfoldAction]:
+    def get_actions_detail(self, request: HttpRequest, object_id: Optional[Any]=None) -> List[UnfoldAction]:
         return self._filter_unfold_actions_by_permissions(
             request, self._get_base_actions_detail()
         )
@@ -445,7 +445,7 @@ class ModelAdmin(ModelAdminMixin, BaseModelAdmin):
         """
         return [self.get_unfold_action(action) for action in self.actions_row or []]
 
-    def get_actions_submit_line(self, request: HttpRequest) -> List[UnfoldAction]:
+    def get_actions_submit_line(self, request: HttpRequest, object_id: Optional[Any]=None) -> List[UnfoldAction]:
         return self._filter_unfold_actions_by_permissions(
             request, self._get_base_actions_submit_line()
         )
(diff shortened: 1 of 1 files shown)


lukasvinclav:
`object_id` is already available in these actions. What is the difference?

https://github.com/unfoldadmin/django-unfold#action-examples

dfg-98:
> `object_id` is already available in these actions. What is the difference?
> 
> https://github.com/unfoldadmin/django-unfold#action-examples

I'm seeking a solution that selectively displays the action (ie. "Save and Publish"). While the object is accessible within the action itself, the action it's currently visible and the way to hide is not based on object/target. 

Ideally, I'd like to prevent the action from appearing altogether for certain conditions of the object (ie published books), rather than relying on validation and error messages within the action.

Here's a breakdown of the current behavior and my goal for my use case with Books:

Current Behavior:

- Action is viewable in all cases.
- Validation within the action triggers a ValidationError for published books.

Desired Behavior:

- Action remains hidden for published books.
- No action visibility for users in this scenario.

Key Questions:

How can we achieve this selective visibility within get_actions_submit_line and get_actions_detail?
Are there alternative approaches that might be more suitable?
I'm open to suggestions and appreciate any guidance!

PR #262, opened 2024-01-18
fix: rm python-semantic-release as dependency

### LINKED ISSUE #261: Can `python-semantic-release` be a dev dependency?
I noticed `python-semantic-release` is listed as a dependency and dev dependency, can this be listed solely as a dev dependency?

kwigley:
resolves #261 if appropriate, please close otherwise! thanks for building this package


lukasvinclav:
Very good but it is missing updated lock file. To get the release as soon as possible, I created #263.

PR #292, opened 2024-02-16
Do not remove selected filters during search

aaalekseev:


lukasvinclav:
Which filters are removed? I think the issue is solved in #293

aaalekseev:
> Which filters are removed? I think the issue is solved in #293

It seems not as currently JS makes straight redirect with single URL param "q" for search query (and clearing all other URL params)

PR #293, opened 2024-02-16
fix: range filters wrong arguments

lukasvinclav:


src/unfold/contrib/filters/admin.py
```diff
@@ -94,15 +97,21 @@ class RangeNumericMixin:
...
             value = params.pop(self.parameter_name + "_to")
-            self.used_parameters[self.parameter_name + "_to"] = value
+            self.used_parameters[self.parameter_name + "_to"] = (
+                value[0] if isinstance(value, list) else value
+            )
 
     def queryset(self, request: HttpRequest, queryset: QuerySet) -> QuerySet:
         filters = {}
 
+        print(self.used_parameters)
```
dominikbullo, line 114: no need for this

PR #343, opened 2024-04-04
fix: unable to set custom rows length for textarea widget

jdenisTLM:
By default, textarea rows is set to 2 in UnfoldAdminTextareaWidget. But it is not possible to change this value, because it is systematicaly updated by the init function of this widget.


lukasvinclav:
@jdenisTLM yes, this will allow to set "rows" attribute for textarea BUT it will not change the default height of the textarea because its height is adjusting based on the content. 

In general, having this feature is okay but we have to find a way how to set approriate min height to hidden div below the textarea which is responsible for auto resizing.

PR #349, opened 2024-04-09
fix: object overwriting when using multiple charts

BitTheByte:
This PR fixes an overwrite bug in the `DEFAULT_CHART_OPTIONS` as it was shared and mutated between chart instances resulting in a wrong rendering.

Before:
image

After:
image


lukasvinclav:
@BitTheByte thank your very much for PR. Would you mind to provide use-case how I can replicate the issue in the screenshot above? I want to test the functionality but I'm not sure where to start.

lukasvinclav:
I was not able to replicate the issue. If you find a time to write proper detailed description how to reproduce, please open new issue linking this PR and I will take a look.

PR #357, opened 2024-04-19
Improve Show As Tabs by Default Class

ElebrUS:
Problematics:
 - Suppose I already have a finished project and I want to connect unfold admin to it.
- All my models are highly loaded and have a lot of fields that I want to group by tabs, specifying classes for each fieldset is a very labor-intensive task

Solution:
- Add default class settings for all fieldsets that can be overridden in one/few models

Features:
 - Added defining a default name for a fieldset if it's a tab and it doesn't have a name assigned to it, i.e. None


lukasvinclav:
I think you can write your own admin mixin which will override `get_fieldsets` method and adds an extra class.

Issue #206, opened 2023-11-28
add fav icon via unfold settings


wiktorcie:
Not available now, should be easy to implement and adds a nice touch.

wiktorcie:
is extrahead of any use to me? Or I need to override default skeleton template?

vic-developer:
no idea, that's what i did.

![image](https://github.com/unfoldadmin/django-unfold/assets/11517372/15586a9a-3be4-432e-a930-a7fdb3daa798)


![image](https://github.com/unfoldadmin/django-unfold/assets/11517372/7cd6476f-d202-4e25-bedb-9c35034cea26)

PR #382, opened 2024-05-12
feat: add site favicon to skeleton.html

furkankurt:
Django Unfold doesn't have a favicon feature.
We had to override the whole `skeleton.html` template to add a favicon. (#206)

I have added a `site_favicon.html` file to the `unfold/helpers` directory, we can add all favicon types to the website.
Instead of overriding the `skeleton.html`, we will override just the `site_favicon.html`.

_PS: I added example favicon links as template comments._

### DIFF
--- a/src/unfold/templates/unfold/helpers/site_favicon.html
+++ b/src/unfold/templates/unfold/helpers/site_favicon.html
@@ -1,0 +1,11 @@
+{% load static %}
+
+{% comment %}
+    This snippet includes the favicon in the head of the HTML document.
+    It is used in the skeleton.html layout template.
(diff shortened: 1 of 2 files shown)


frobert:
I would suggest you also add option to change the favicon in settings.py UNFOLD config the same way as SITE_ICON, ex.

`"FAVICON": {
        "light": lambda request: static("favicon.png"),
        "dark": lambda request: static("favicon_dark.png"),
    },`

furkankurt:
> Hey, thanks for good initial PR. Would you mind to create an issue for this task? Don't forget to link this PR there.

Hello, thank you for your valuable comment. I've created an issue about this PR.

PR #402, opened 2024-05-21
fix: django-import-export unable to import

ZairMahmood:
## Issue
Import action is broken per changes from #398 

Resource was inferred as the file field but it's a different field on its own. Current issue can also be observed on the deployed [unfoldadmin/formula](https://github.com/unfoldadmin/formula)
The actual file is never passed back to import-export since the expected input_file field is missing.

## Changes
Reverting partial changes from #398 while cleaning up the formatting a touch to manage multiple resources

Single Resource View:
Screenshot 2024-05-21 at 3 44 50 PM

Multiple Resource View:
Screenshot 2024-05-21 at 3 45 05 PM

## Tests
Tested locally with:
django-unfold@0.24
django@5
django-import-export@4.0.3

I didn't go too deep in the export action and how resources might play a part there since it was functional and not completely broken.

### DIFF
--- a/src/unfold/contrib/import_export/forms.py
+++ b/src/unfold/contrib/import_export/forms.py
@@ -1,15 +1,29 @@
-from import_export.forms import ImportExportFormBase as BaseImportExportFormBase
-from unfold.widgets import SELECT_CLASSES, UnfoldAdminFileFieldWidget
+from import_export.forms import ExportForm as ExportFormBase
+from import_export.forms import ImportForm as ImportFormBase
+from unfold.widgets import UnfoldAdminFileFieldWidget, UnfoldAdminSelectWidget
 
 
-class ImportForm(BaseImportExportFormBase):
+class ImportForm(ImportFormBase):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.fields["resource"].widget = UnfoldAdminFileFieldWidget()
-        self.fields["format"].widget.attrs["class"] = " ".join(SELECT_CLASSES)
+        resource_widget = self.fields["resource"].widget
+        if len(resource_widget.choices) > 0:
+            self.fields["resource"].widget = UnfoldAdminSelectWidget(
+                attrs=resource_widget.attrs, choices=resource_widget.choices
+            )
+        else:
+            self.fields.pop("resource")
+        self.fields["import_file"].widget = UnfoldAdminFileFieldWidget()
+        format_widget = self.fields["format"].widget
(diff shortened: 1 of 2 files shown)


src/unfold/contrib/import_export/templates/admin/import_export/import_form.html
```diff
@@ -4,26 +4,28 @@
     <form action="" method="post" enctype="multipart/form-data">
         {% csrf_token %}
 
-        <p class="bg-blue-50 mb-8 text-blue-500 px-3 py-3 rounded-md text-sm dark:bg-blue-500/20 dark:border-blue-500/10">
+        <div class="bg-blue-50 mb-8 text-blue-500 px-3 py-3 rounded-md text-sm dark:bg-blue-500/20 dark:border-blue-500/10">
```
ZairMahmood, line 7: Since `dl` tags can't be inside `p` tag, switched to `div` to capture implemented classes

src/unfold/contrib/import_export/forms.py
```diff
@@ -1,15 +1,29 @@
...
 
-class ImportForm(BaseImportExportFormBase):
+class ImportForm(ImportFormBase):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.fields["resource"].widget = UnfoldAdminFileFieldWidget()
-        self.fields["format"].widget.attrs["class"] = " ".join(SELECT_CLASSES)
+        resource_widget = self.fields["resource"].widget
+        if len(resource_widget.choices) > 0:
+            self.fields["resource"].widget = UnfoldAdminSelectWidget(
```
ZairMahmood, line 11: Replaced to using widgets here since they're already part of the unfold project and don't have to rewrite the same logic.
Not sure if there was a proper way of using this widget but this worked instead of just `UnfoldAdminSelectWidget`

ZairMahmood:
Another suggestion for future is that the `UnfoldAdminFileFieldWidget` can allow upload from clicking anywhere inside of the box instead just the little icon off to the side.

ZairMahmood:
I didn't go thru each template from import_export but possibly needs some updates/restyling there also to match updates from their v4

lukasvinclav:
First of all, thank you very much for high quality feedback. Based on your code, I created new PR fixing multiple issues. If you have time, would you mind to test my PR and provide another feedback?

https://github.com/unfoldadmin/django-unfold/pull/404/files

lukasvinclav:
Closing in favor of #404. Thank you very much for valuable feedback.

PR #404, opened 2024-05-22
fix: import export 4 support

lukasvinclav:


ZairMahmood:
Import and export work. The formatting for multiple resources look like this:

image

One tiny bug I found: 
- If you select a file
- Import it but don't confirm.
- Go back a page (using browser back button)
- Then select a different file again
- The file field doesn't update with the file name.

ZairMahmood:
otherwise, lgtm since import is working again 👍

lukasvinclav:
Fixed the message. I was not able to replicate the issue with filenames. In the future we can open a new ticket for it.

PR #453, opened 2024-06-07
fix: Prevent mutating navigation and tab settings

vinayinvicible:
If the `link` is a callable, it is getting evaluated only once. The result of this evaluation is being updated in the settings object.

Because of this, all users get the same link instead of it getting evaluated per request.

cc: @lukasvinclav


vinayinvicible:
@lukasvinclav Can you please have a look at this

lukasvinclav:
Fixed in #540

PR #474, opened 2024-06-12
fix: sidebar dark mode

lukasvinclav:


VaZark:
I think the latest version 0.28.1 broke the theming

image

PR #547, opened 2024-07-07
feat: choices support in ArrayWidget

lukasvinclav:


Tonkonozhenko:
Hi @lukasvinclav 
I tried using the feature, but I don't understand how to allow adding multiple values.

```
# model.py
...
also_evaluate_for: list[str] = ArrayField(models.CharField(), choices=PeriodTypes, blank=True, null=True)
```

result:
image

PR #549, opened 2024-07-08
feat: inline tabs

lukasvinclav:


frobert:
Not sure this is a good option, if this is correct, now we have another tab group above the main fieldsets tab group?

It would be a lot better to have the option to add inlines inside model admin fieldsets config and have the inline in the same tab group. This is a good example of inlines within the fieldsets: https://github.com/robertkovac/django-fieldsets-with-inlines but thisone does not work in unfold.

Screenshot 2024-07-20 at 13 05 47

PR #602, opened 2024-07-19
add 127.0.0.1 to ALLOWED_HOSTS

AliYmn:


AliYmn:
@lukasvinclav I added that it is generally preferred to work from http://127.0.0.1:8000/admin I don't think localhost/admin is used much.

lukasvinclav:
are you using devcontainer in vscode to run the server?

lukasvinclav:
Thanks for the PR but from my perspective there is nothing broken and it works as expected.

PR #603, opened 2024-07-19
docs: add Chart.js credits

raayu83:
Took me a while to figure out what is used for Charts since there was no reference.
I've added the info to the README.md in this PR.


lukasvinclav:
Thanks for contribution.

PR #623, opened 2024-07-24
docs: missing group import

sabbir2609:
need to add `from django.contrib.auth.models import Group`


lukasvinclav:
Thanks for PR:

1. why the Group model has a separate line? It can be imported on the same line as User
2. please update the title to something meaningful

lukasvinclav:
Thanks for contribution

Issue #624, opened 2024-07-24
A model with primary key type UUID with actions_row helper will raise JS error


galetahub:
**What version of Unfold are you using?**

0.33.0
0.32.0

**What version of Django are you using?**

5.0.6

**What browser are you using?**

Arc 
Version 1.51.0 (51691)
Chromium Engine Version 126.0.6478.127

**Did you checked changelog/commit history, if the bug is not already fixed?**

Yes

**Did you searched other issues, if the bug is not already fixed?**

No

**Describe your issue**

The main issue is that Alpine.js does not support "-" in variable names. 
But when we use UUID our model pk has "-" symbols.

When using actions_row helper with UUID model will raise JS error:
```
alpine.js:1 Alpine Expression Error: Invalid or unexpected token
```

Code example:
```python
# models.py
from django.db import models
from django.contrib.postgres.functions import RandomUUID

class Post(models.Model):
    id = models.UUIDField(primary_key=True, db_default=RandomUUID(), editable=False)
    title = models.CharField(max_length=255)
```

```python
# admin.py
from unfold.admin import ModelAdmin
from .models import Post

@admin.register(Post)
class PostAdmin(ModelAdmin):
   list_display = ("id", "title",)
   actions_row = [
        "show_post_action",
        "edit_post_action",
    ]
    # ...
```
<img width="1712" alt="SCR-20240724-qyfs" src="https://github.com/user-attachments/assets/5510e9b6-22b4-4f2c-86da-df6c1e9e5a62">

<img width="707" alt="SCR-20240724-qzxi" src="https://github.com/user-attachments/assets/fbebdaf7-2874-4492-9e73-9ecf087e1a65">

lukasvinclav:
Thank you very much for report and PR which I already merged. It seems that x-ref in alpine.js does not support hyphens.

PR #625, opened 2024-07-24
fix: remove hyphen from instance_pk (UUID) to fix issue #624

galetahub:
Fix JS error for models with primary key type UUID:

```
alpine.js:1 Alpine Expression Error: Invalid or unexpected token
```


lukasvinclav:
Thank you for your contribution.

Issue #613, opened 2024-07-22
 Custom HTML attributes for forms and fields (form-row) 


juanfcocontreras:
**Do you have any examples?**

https://alpinejs.dev/essentials/templating#toggling-elements

**Is it a third party application feature request?**

For example: No

**In case of third party application, when it was updated last time?**

Not applicable

**Describe your feature request**

It would be interesting to have the ability to add custom attributes at both form and field level (including their title, i.e. form-row). This would allow, in a very simple way and without having to overwrite templates (which have the inconvenience of having to revise after updates), to add Alpine.js capabilities such as showing or hiding fields depending on the values you have in other fields (x-show).

I think you could use an approximation similar to that of:

https://github.com/unfoldadmin/django-unfold/pull/11/files

Do you think it would be useful and could be implemented?

galetahub:
A simple example, how to add custom attributes to form fields:
```python
from django import forms

class PostForm(forms.ModelForm):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # checkbox
        self.fields["public_author"].widget.attrs.update({
            'x-ref': 'public_author_checkbox',
            'x-on:change': '$refs.author_name_input.disabled = !$refs.public_author_checkbox.checked'
        })
        # input
        self.fields["author_name"].widget.attrs.update({
            'x-ref': 'author_name_input',
            'x-bind:disabled': '!$refs.public_author_checkbox.checked'
        })
```

Finally connect form with admin model:
```python
from django.contrib import admin
from unfold.admin import ModelAdmin

@admin.register(Post)
class PostAdmin(ModelAdmin):
    list_display = ('id', "title", "author_name")
    form = PostForm
```

juanfcocontreras:
But this would not include the labels, only the fields, or am I wrong?

> A simple example, how to add custom attributes to form fields:
> 
> ```python
> from django import forms
> 
> class PostForm(forms.ModelForm):
>     def __init__(self, *args, **kwargs):
>         super().__init__(*args, **kwargs)
>         # checkbox
>         self.fields["public_author"].widget.attrs.update({
>             'x-ref': 'public_author_checkbox',
>             'x-on:change': '$refs.author_name_input.disabled = !$refs.public_author_checkbox.checked'
>         })
>         # input
>         self.fields["author_name"].widget.attrs.update({
>             'x-ref': 'author_name_input',
>             'x-bind:disabled': '!$refs.public_author_checkbox.checked'
[...]

PR #631, opened 2024-07-26
feat: add pre/post-form and pre/post-form_row blocks

juanfcocontreras:
Adds new blocks in the templates for `form` and `form-row` that allow to use Alpine.js functionalities in a more comfortable way.

More details available in:

https://github.com/unfoldadmin/django-unfold/issues/613#issuecomment-2248208039


lukasvinclav:
Thanks for the PR but at the moment there are no plans adding these new blocks. I recommend overriding templates to fit your use case.

juanfcocontreras:
First of all, I want to express my gratitude for your hard work on django-unfold and for sharing it with the community. It's an amazing project that I use daily. This is why I've always been eager to contribute to its improvement, even with my limited Django knowledge.

I understand that this is your project, and I fully respect the decisions you make regarding it. However, I must admit that I was disappointed to see the cancellation of my pull request.

After our conversation, creating a screencast, summarizing key changes, and uploading a minimal project to demonstrate the new functionality and receiving the support of other users, it was disheartening to receive a response that felt quite brief, like "thanks but no thanks"

I really want to avoid overriding templates because they change with each version, and maintaining them can be a hassle. Extending blocks seems to be a practical solution, especially since I'm currently stuck on version 0.20 of django-unfold for my ongoing project.

What I find most challenging is not just the non-acceptance of the change, which I understand may not suit the current direction, but the lack of a technical explanation for the decision. It’s clear that the proposed changes had positive aspects.

I hope my message comes across clearly, English is not my first language. I genuinely appreciate your project, but I find it frustrating to try contributing and not understanding where I might have gone wrong when my efforts are not accepted.

lukasvinclav:
First of all, thank you very much for your kind words; they mean a lot to me. Regarding your feedback:

You had to create a screencast and code example because your feature is not very common, and, at least for me, it was difficult to imagine what the final result should be. In the original issue, you can see that I was really trying to understand what you wanted to achieve.

It's important to note that I never promised your PR would be merged. I still believe that even if a PR is closed without being merged, it can be beneficial for others who might reuse your example, so the work won't be wasted.

In general, you have several options for getting your business logic to work:

- You can override the template. This is an extremely simple solution that literally takes a few minutes.
- You can write your own JS file to handle the business logic. Again, this is very simple, and you don't have to override anything. Unfold supports including JS files by default.
- If you're concerned that future releases of Unfold might break your code, you can fork the repository and introduce your changes there.

As you can see, there are multiple ways to incorporate your changes into your project without shifting your tasks onto the project maintainer.

The reason your code was not merged is that it is not on the current roadmap. I believe your change is something you can implement yourself in under a minute, so there’s no reason to include it in Unfold by default, as it would increase the project's complexity. I'm not saying this can't change in the future, but today is not the right time to include it.

It's important to recognize how complex the project already is and how challenging it is to maintain it properly while still delivering new features and fixes. On a regular basis, in 95% of situations, I have to say "no" to a lot of ideas, even if I like them, to keep the project running smoothly and optimize my time.

Thank you very much for your understanding.

juanfcocontreras:
Thank you very much for your kind reply. I fully understand your point of view and respect it completely.

I would just like to make a few points and ask you for advice for one of your approaches to get the logic I want.

First of all I think that the fact that I had to make a screencast was rather due to a lack of command of English. I think my use case is not that rare, and proof of that is that several users voted for my pull request.

In fact, what I want is just what is in the screencast, to be able to hide or show fields depending on others (not just disable). It's a very common thing in forms, and in fact as they say in Alpine.js the directive they use for it is one of the most used in their framework (https://alpinejs.dev/directives/show).

As for the approaches you propose, the first one is the one I wanted to rule out with the pull request. To give you an example, the template `src/unfold/templates/admin/change_form.html` has changed 3 times in the last 9 versions of django-unfold. This is exactly what I want to avoid.

As for the second approach, I find it very interesting, but I don't know if you can guide me. The idea of doing it like in the pull request is to try to take advantage of Alpine.js functionalities, in particular `x-show`. But I don't see capturing that logic with Javascript as long as I use Alpine.js (and don't have to rely to jQuery). Do you know where I can find an example of this approach? You say it's very simple, so I guess it shouldn't be difficult to find one.

As for the third approach, I don't think about it at all. You know your project much better than I do, and if you consider that this change increases its complexity, I certainly trust your judgement much more than mine in this case. The idea of making a fork outside the developer makes no sense when the developer is friendly, justifies his answers and is willing to argue for them, as you are. I don't consider it at all.

Thank you very much again for your project and attitude.

Issue #614, opened 2024-07-22
"Save as" button width glitch in admin form


gleniat:
unfold 0.32.0
django 5.0.7
latest Chrome

**Did you searched other issues, if the bug is not already fixed?**

Yes

**Describe your issue**

If you activate the 'Save As' button in the admin form, you will notice that the button is too wide. Removing the 'w-full' class seems to fix this issue. The 'w-full' class is also applied to other buttons, but I'm not sure if it is intended.

![unfold-save-as-button](https://github.com/user-attachments/assets/99f20f90-738c-427d-b41f-53c8d66bd4b1)

```
from django.contrib import admin
from testapp.models import Person

class PersonAdmin(admin.ModelAdmin):
    save_as = True

admin.site.register(Person, PersonAdmin)
```

lukasvinclav:
Completed in #668

PR #637, opened 2024-07-28
Fix: 'Save as' button width glitch in admin form.

khaled5321:
Fix issue: #614

Screenshot:
Screenshot 2024-07-28 134832


lukasvinclav:
Thanks for pointing me at the issue but the correct solution is #668

PR #674, opened 2024-08-12
feat: ensure user has access to any item in sidebar group

Tonkonozhenko:
PR also fixes wrong nesting in template


Tonkonozhenko:
@lukasvinclav could you review the PR please?

lukasvinclav:
Thanks for PR. I found simpler solution here #719. In general, theme will hide all groups which are not containing any links just by using pure CSS.

PR #682, opened 2024-08-15
feat: tabs with URL navigation

### LINKED ISSUE #570: Remember/show current active TAB (anchor tag link)
Hi there, when working wuth admin, I've noticed active TAB is not remembered in admin. 

It would be nice to have working link anchor tags, so when going to for example http://localhost:8000/contacts/company/1/change/#TabName the view would open the specified TAB called "TabName" directly.

Thanks, best regards

marcpicaud:
Fixes #570

Hey there,

First, thank you for django-unfold, I love it and I'm pretty sure I'm not alone.

Saving the `activeTab` state would benefit the users IMO: 
Refreshing the page, sharing a link, navigating back and forth would be possible without losing the tab state.

In this PR:
* Save the `activeTab` state in the URL when clicking tabs
* Initialize the `activeTab` variable from the URL during a full reload

There's one problem with my current approach : 
If any hash is detected in the URL, it will become the `activeTab` value. If the hash doesn't match with any tab href, there will be no tab displayed. Maybe to fix this we should have a naming convention for tabs ? Let me know what you think !

Cheers

### DIFF
--- a/src/unfold/templates/unfold/helpers/tab_list.html
+++ b/src/unfold/templates/unfold/helpers/tab_list.html
@@ -18,6 +18,7 @@
                     {% if inlines_list %}
                         <li class="border-b last:border-b-0 md:border-b-0 md:mr-8 dark:border-gray-800">
                             <a class="block cursor-pointer font-medium px-3 py-2 md:py-4 md:px-0"
+                               href="#general"
                                x-on:click="activeTab = 'general'"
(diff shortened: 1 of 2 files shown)


lukasvinclav:
Awesome work! Thanks for contribution.

frobert:
I don't think this is working in latest version... Tabs in change view do not set the # in the url when clicking...

lukasvinclav:
@frobert if you found a bug, file a new issue here on GitHub

Issue #694, opened 2024-08-21
URLField widget: display a clickable link when a value is set


marcpicaud:
**Is it a third party application feature request?**
No

**Describe your feature request**

URLField widget could display an actual link

**Do you have any examples?**

Native Django admin : 
<img width="788" alt="image" src="https://github.com/user-attachments/assets/89ced4b7-bb75-47a5-b0c0-d3c5e1b6f93e">

khaled5321:
Hi @lukasvinclav, I have created a draft PR for this issue #723. Can you review it?

PR #723, opened 2024-09-07
feat: URLField widget, display a clickable link

khaled5321:
related to: #694

Add a Widget for the URL field that displays a clickable link when the URL is valid.

Screenshot 2024-09-07 144132


lukasvinclav:
Thanks for contribution.

PR #731, opened 2024-09-10
Fix file field upload

radzih:
[Fix file input[type=text] selector](https://github.com/unfoldadmin/django-unfold/commit/1f9d8f55454c14bd48bc35dea00e39eaaff39789)


lukasvinclav:
Fixed in #716

PR #743, opened 2024-09-15
fix: add missing error messages to login template

jose-padin:
There were missing error messages when a user tried to perform a log in.
In this case, the errors came from the [django-password-expire](https://pypi.org/project/django-password-expire/) package. Some users were blocked because their passwords had expired, but they couldn´t see the error in the UI.
This PR fixes this bug by adding those error messages to the src/unfold/templates/admin/login.html template.

### DIFF
--- a/src/unfold/templates/admin/login.html
+++ b/src/unfold/templates/admin/login.html
@@ -36,6 +36,12 @@
 
                 {% include "unfold/helpers/messages/error.html" with errors=form.non_field_errors %}
 
+                {% if messages %}
+                    {% for message in messages %}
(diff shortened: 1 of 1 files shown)


lukasvinclav:
Why you closed the issue which you created for this PR? Does it mean that you already solved your problem and you just forgot to close this PR as well?

jose-padin:
Hi, @lukasvinclav. I didn't close the issue, I closed a previous PR because I thought it was not well done. I didn't create an issue for this bug.  Do you want me to create one?

lukasvinclav:
Thanks for report. Issue fixed in this PR  #772

PR #750, opened 2024-09-17
chore: update material symbols

PavelPancocha:
There are already new symbols/icons added to Material Symbols, eg. `id_card` which were not included in Unfold static file.

I added the file version to the file name, making it easier to find which version is supported.


lukasvinclav:
Please create an issue for this PR with the link pointing at the site where you downloaded new icon set. I will take a look at it and I will manually update fonts to be sure that everything is okay. 

Now, I'm going to close this PR, not because it is bad but because I have to be sure that we are using right fonts. The red flag here for me is that there is almost 100KB difference between new and old font and I would be expecting that new font is going to be bigger but it is actually smaller what is not making sense to me.

PR #752, opened 2024-09-18
fix: form signature

PavelPancocha:
- Fix wrong typing
 - Improve code

Fix: #760

### DIFF
--- a/src/unfold/forms.py
+++ b/src/unfold/forms.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Iterable, Optional, Tuple
 
 from django import forms
 from django.contrib.admin.forms import (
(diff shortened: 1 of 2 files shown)


src/unfold/forms.py
```diff
@@ -92,41 +93,54 @@ def __init__(
 class UserChangeForm(BaseUserChangeForm):
     def __init__(
         self,
-        request: Optional[HttpRequest] = None,
         *args,
         **kwargs,
     ) -> None:
-        super().__init__(request, *args, **kwargs)
+        super().__init__(*args, **kwargs)
```
PavelPancocha, line 99: There is no "request" in this form. Only data as first argument. Why it worked fine: Djano uses it as a positional only -> the "request" was "data"

lukasvinclav:
Would you mind to create appropriate issue for this ticket explaing in detail what we want to achieve?

PavelPancocha:
@lukasvinclav here you go: #760

PR #785, opened 2024-10-06
feat: sortable inlines

lukasvinclav:


Zerotask:
I just tested it and it works really good. Nice and clean.
Is there any chance to add this for ModelAdmin as well?

lukasvinclav:
hey, thanks for testing Patrick. I didn't implemented changelist sorting because it will be much more complex than sorting of inlines. To get everything done correctly will consume a lot of my spare time so there is no ETA for functioniality.

PR #787, opened 2024-10-06
feat: add multiplerelateddropdownfilter with cache cupport

aozeel:
Add multiplerelateddropdowncachefilter to reduce response time of admin list pages of huge related filter tables


lukasvinclav:
I think the functionality might be good but isn't it better to put your custom caching into your project directly? You can extend the filter by your own business logic in your project and then just reuse new class.

PR #792, opened 2024-10-07
Fix django-guardian permission display

### LINKED ISSUE #791: Fix django-guardian permission display
The `django-guardian` permissions are rendered in reverse.

Screenshot 2024-10-07 at 4 01 38 PM
Screenshot 2024-10-07 at 4 02 08 PM
Screenshot 2024-10-07 at 4 02 00 PM

Without unfold
Screenshot 2024-10-07 at 4 05 35 PM

gmargaritis:
Resolves #791


lukasvinclav:
Thanks for the fix. I addressed the issue in another PR #793 because:

1.  I didn't reverted first if by using "not" to keep same flow as it is in guardian

2. I found another issue in user table which I fixed in my PR as well

gmargaritis:
LGTM I had accidentally left a `not` in there

PR #811, opened 2024-10-18
Enhance Tab Functionality for URL Query-Based Field Filtering

Ehco1996:
## Summary:

This PR introduces improvements to the tab functionality, allowing tabs to serve as quick filters for field data based on URL query parameters. The current logic does not handle this scenario correctly, leading to incorrect filtering results.

## Example:

Currently, the tabs are structured as follows:
```
tabs = [
    {
        "models": [
            "group1",
        ],
        "items": [
            {"title": "group-1", "link": "/admin/app_name/group/"},
            {"title": "group--2", "link": "/admin/app_name/group/?lable=online"},
        ],
    },
]
```
This PR addresses the issue by implementing a more robust logic to correctly interpret and apply the URL query parameters for filtering.

### DIFF
--- a/src/unfold/sites.py
+++ b/src/unfold/sites.py
@@ -1,5 +1,6 @@
 from http import HTTPStatus
 from typing import Any, Callable, Dict, List, Optional, Union
+from urllib.parse import parse_qs, urlparse
 
 from django.contrib.admin import AdminSite
 from django.contrib.auth import REDIRECT_FIELD_NAME
@@ -394,10 +395,25 @@ class UnfoldAdminSite(AdminSite):
     def _get_is_active(self, request: HttpRequest, link: str) -> bool:
         if not isinstance(link, str):
             link = str(link)
-
-        if link in request.path and link != reverse_lazy(f"{self.name}:index"):
+        request_path = request.get_full_path()
+        # Special case for the index page
+        if link == request.path == reverse_lazy(f"{self.name}:index"):
             return True
-        elif link == request.path == reverse_lazy(f"{self.name}:index"):
-            return True
 
-        return False
+        link_parts = urlparse(link)
+        request_parts = urlparse(request_path)
+
+        # Check if paths match
+        if link_parts.path != request_parts.path:
+            return False
+        # If there are no query parameters in the link, it's active only if the request also has no query parameters
+        if not link_parts.query:
+            return not request_parts.query
+        # Parse query parameters
+        link_params = parse_qs(link_parts.query)
+        request_params = parse_qs(request_parts.query)
+        # Check if all link parameters are present in the request with the same values
+        return all(
+            key in request_params and request_params[key] == value
+            for key, value in link_params.items()
+        )


lukasvinclav:
Would you mind to upload a short video what does this functionality do?

Ehco1996:
here is the gif,please note that all tab refer to the same model wiith different filter paramters as i mentioned in PR description

pnost:
Nice. I used it for a link in settings.py -> SIDEBAR -> navigation:
`"link": lambda request: reverse_lazy("admin:rezeptor_prescriptionsync_changelist")
                        + "?kassenrezept=true&status__exact=processing`
which now does not collapse anymore but stays open & active. Thank you

lukasvinclav:
I think, the @pnost solution is correct. You don't need to change any business logic to get tabs with filters.

pnost:
@lukasvinclav it's not only about having tabs with filters but having them hightlighted as **active** tab. 
Links in tabs and navigation are not highlighted as active until now if you use the same model but add a filter in the url. @Ehco1996 commit changes this.

Ehco1996:
@lukasvinclav ptal

subham1099:
The PR doesn't work when the tabs have completely different URL/model - it doesn't highlight the "Locations" tab correctly when filters are applied.

Without filters: [screenshot](https://github.com/user-attachments/assets/ec57c933-82e7-4f1f-9f2c-17367012668a)
with filters: [screenshot](https://github.com/user-attachments/assets/41e06f9c-c5d8-4540-98ea-9ed14e0b5cb5)

My use case is I want to show a custom page "Dashboard" in the tab group while also showing the tabs, so I had to pass the "cl" of "Locations" model inside the dashboard page context, like below:

```py
from django.urls import path
from django.views.generic import TemplateView
from unfold.views import UnfoldModelAdminViewMixin
from unfold.admin import ModelAdmin

class MyClassBasedView(UnfoldModelAdminViewMixin, TemplateView):
    model_admin: ModelAdmin

    title = "Locations Dashboard"  # required: custom page header title
    permission_required = ()  # required: tuple of permissions
    template_name = "admin/app/location/dashboard.html"

    def get_context_data(self, **kwargs: t.Any) -> dict[str, t.Any]:
        context = super().get_context_data(**kwargs)
        cl = self.model_admin.get_changelist_instance(self.request)
        return {**context, "cl": cl, "opts": cl.opts}

class LocationAdmin(ModelAdmin):
    ...


    def get_urls(self) -> list[URLPattern]:
        return [
            *super().get_urls(),
            path(
                "dashboard",
                MyClassBasedView.as_view(
                    model_admin=self
                ),  # IMPORTANT: model_admin is required
                name="app_generallocation_dashboard",
            ),
        ]
```

if you notice, the url for dashboard is `/app/generallocation/dashboard` and the code in the main branch highlights both "Dashboard" and "Locations" tab because it compares using the "in" operator - [screenshot](https://github.com/user-attachments/assets/6d8f99a7-4fa9-46f2-9621-b7f6321b6665).

That is when I found this PR. This PR works, but breaks when filters are applied (see screenshots above).

lukasvinclav:
Closing this PR in favor of #876

PR #824, opened 2024-10-23
Adding GeneratedField to supported types for RangeNumericFilter

yarimiz:
I have a use case where I wish to have a RangeNumericFilter for a GeneratedField.
This was not allowed until now. This PR is to allow this configuration.

I had not added any extra validations to check whether the actual value of the GeneratedField is numeric, as I don't think this should be done by unfold. Let me know if you think otherwise.

### DIFF
--- a/src/unfold/contrib/filters/admin.py
+++ b/src/unfold/contrib/filters/admin.py
@@ -14,7 +14,6 @@ from django.db.models.fields import (
     FloatField,
     IntegerField,
 )
-from django.db.models.fields.generated import GeneratedField
 from django.forms import ValidationError
 from django.http import HttpRequest
 from django.utils.dateparse import parse_datetime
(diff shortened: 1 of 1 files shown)


lukasvinclav:
What type is GeneratedField in database?

Zerotask:
> What type is GeneratedField in database?

depends on what you pass for `output_field`. For example `output_field=models.CharField(max_length=16)` would be a VARCHAR(16)

lukasvinclav:
if the column type is variable, how we are checking if this field can be used for numeric filtering?

yarimiz:
One way we can resolve this is to make sure the output_field of the GeneratedField is one of the already supported field types (DecimalField, IntegerField, FloatField, AutoField)

lukasvinclav:
exactly, if you can, please update the PR

yarimiz:
Sure, working on that now

lukasvinclav:
please ensure that your change is going to be backward compatible as well. I guess on django pre 5.x import fill fail.

yarimiz:
what is the best appraoch you suggest for safe importing? I thought either importing only if django is >=5.0 at the top level, or do the whole importing the GeneratedField.output_field checks inside a try-except and handling the ImportError if raised.

lukasvinclav:
In the past I used the approach below but if you are aware of something more elegant, I'm open to incorporate it everywhere:

https://github.com/unfoldadmin/django-unfold/blob/main/src/unfold/admin.py#L59-L65

yarimiz:
I don't think my appraoch is more elegant, but just might be more "local"

```
try:
    from django.db.models import GeneratedField
     if isinstance(field, GeneratedField) and isinstance(
        field.output_field, (DecimalField, IntegerField, FloatField, AutoField)
    ):
        raise TypeError(
            f"Class {type(field.output_field)} of the output_field is not supported for {self.__class__.__name__}."
        )
except ImportError:
    pass
```

Tell me which you prefer and I'll implement it

lukasvinclav:
Your snippet looks good as well.

yarimiz:
Please take a look at the final code when you have the time :)

lukasvinclav:
I played with simpler check for backward compatibility. What do you think about the solution below. Would you mind to test it?

```python
if not isinstance(
    field.output_field if hasattr(field, "output_field") else field,
    (DecimalField, IntegerField, FloatField, AutoField),
):
    raise TypeError(
        f"Class {type(self.field)} is not supported for {self.__class__.__name__}."
    )
```

Zerotask:
I noticed something related to GeneratedFields. I'm not sure if you want to include that in this PR.
There are probably some more GeneratedFields related issues.

This one here is treated slightly differently than a regular `BooleanField`:
```python
is_successful = models.GeneratedField(
    expression=models.Q(http_status_code__range=(200, 299)),
    output_field=models.BooleanField(),
    db_persist=True,
)
```

image

image

---

a regular BooleanField looks like this:
image

image

(the green background color is customization which is visible, because the screenshot function is too slow :D)

lukasvinclav:
I'm closing this pull request due to inactivity.

PR #854, opened 2024-11-15
feat: language switch

lukasvinclav:


Zerotask:
Nice addition! I just tested it and got the error
> Reverse for 'set_language' not found. 'set_language' is not a valid view function or pattern name.

> `<form action="{% url 'set_language' %}" method="post" class="flex w-full">`

We do have 5 languages, which are also defined in the django settings, but we have our own view and don't use `set_language` (maybe we reconsider it in the future).
Therefore it would be a good idea to check, if it exists and/or add the info in the settings docs.

lukasvinclav:
@Zerotask If you are not using `set_language` then don't enable the language switch because it will not work.

elias-ebrahimpour:
Dear @lukasvinclav, thanks again for accepting the pull request! I previously mentioned adding my name to the contributors' list, and I wanted to kindly remind you about it now that the pull request has been merged if it's possible. I’d appreciate it!

PR #856, opened 2024-11-15
fix: non-dynamically added filefields

### LINKED ISSUE #855: `fileInputUpdatePath` stopped applying to exisiting file inputs (introduced in 0.41.0)
*Versions:* Django 5 + Unfold 0.41 + latest Chrome

It looks like no. I am not sure that the demo site has 0.41.0 yet

## Description
There's an issue with the `fileInputUpdatePath` implementation that prevents it from working with file inputs that exist when the page initially loads. This was introduced in #846.

Currently, the MutationObserver only watches for newly added elements (`childList` mutations) and attaches the change event handler to file inputs after they're dynamically added to the DOM. However, this means any file inputs that are part of the initial page HTML are missed completely.

### Suggested Fix

We should modify the function to:
1. Immediately attach handlers to existing file inputs when the function is called
2. Keep the MutationObserver for handling dynamically added inputs

Here's how the code could be restructured:
```javascript
const fileInputUpdatePath = () => {
  const handleFileInput = (input) => {
    input.addEventListener("change", (e) => {
      const parts = e.target.value.split("\\");
      const placeholder = input.parentNode.parentNode.parentNode.querySelector(
        "input[type=text]"
      );
      placeholder.setAttribute("value", parts[parts.length - 1]);
    });
  };

  // Handle existing inputs
  document.querySelectorAll("input[type=file]").forEach(handleFileInput);

  // Watch for new inputs
  const observer = new MutationObserver((mutations) => {
    for (const mutation of mutations) {
      if (mutation.type === "childList") {
        document.querySelectorAll("input[type=file]").forEach(handleFileInput);
      }
    }
  });

  observer.observe(document.body, {
    childList: true,
    subtree: true,
  });
};
```

yardensachs:
Fixes #855

### DIFF
--- a/src/unfold/static/unfold/js/app.js
+++ b/src/unfold/static/unfold/js/app.js
@@ -135,19 +135,24 @@ const dateTimeShortcutsOverlay = () => {
  * File upload path
  *************************************************************/
 const fileInputUpdatePath = () => {
+  const handleFileInput = (input) => {
+    input.addEventListener("change", (e) => {
+      const parts = e.target.value.split("\\");
+      const placeholder = input.parentNode.parentNode.parentNode.querySelector(
+        "input[type=text]"
+      );
+      placeholder.setAttribute("value", parts[parts.length - 1]);
+    });
+  };
+
+  // Handle existing inputs
+  document.querySelectorAll("input[type=file]").forEach(handleFileInput);
+
+  // Watch for new inputs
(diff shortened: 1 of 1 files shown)


lukasvinclav:
Hey, thanks for PR. I submitted new PR #861 handling one more edge case to prevent binding `change` event multiple times.

Issue #839, opened 2024-11-06
Capturenode doesn't define child_nodelist


Mte90:
So if you use django-fastdev with django-unfold the admin dashboard is not accessible.

I reported the issue on https://github.com/boxed/django-fastdev/issues/51 also with a patch but the package author says that the `Capturenode` doens't define child_nodelist.

Ref: https://github.com/unfoldadmin/django-unfold/blob/541a1577ec93630a9c99ec623d395a26c7098b65/src/unfold/templatetags/unfold.py#L79

lukasvinclav:
Please submit a PR because I have no experience with mentioned third party app.

PR #863, opened 2024-11-18
fix(template): capturenode chidlist

Mte90:
Ref: https://github.com/unfoldadmin/django-unfold/issues/839


lukasvinclav:
Below you can see that `CaptureNode` is inheriting from `Node` which already implements `child_nodelists = ("nodelist",)`. I tested if the value is set in `__init__` and it worked. Are we sure that the PR is really fixing your issue?

https://github.com/unfoldadmin/django-unfold/blob/main/src/unfold/templatetags/unfold.py#L79

https://github.com/django/django/blob/c56e1273a9d87ffad3a84fb597550f79b9820281/django/template/base.py#L960

Issue #873, opened 2024-11-20
Date hiearchy is reset on dropdown filter apply


PavelPancocha:
Relates to: https://github.com/unfoldadmin/django-unfold/issues/868

**What version of Unfold are you using?**
0.41.0

**What version of Django are you using?**
4.2/5.1

**Did you checked changelog/commit history, if the bug is not already fixed?**
Yes

**Did you searched other issues, if the bug is not already fixed?**
Yes

**Did you checked documentation?**
Yes 

**Are you able to replicate the bug in the demo site?**
No, there is no page with date-hierarchy implemented (maybe add it?)

**Describe your issue**

1. You are on changelist of admin with date hierarchy enabled + dropdown filters
2. You select date hiearchy
3. You apply filter
4. Date hierarchy is lost
See: https://www.loom.com/share/3c893a6a858f4dcf9a10515a86c78ff0?sid=523a53a7-2d52-48ba-b818-6cc0e1885047

lukasvinclav:
@PavelPancocha date hierarchy example added here:

https://demo.unfoldadmin.com/en/admin/formula/race/

PR #875, opened 2024-11-22
fix: date hierarchy is reset on filter (#873)

PavelPancocha:
fix: #873


lukasvinclav:
@PavelPancocha please pull latest changes and let me know if it works on your side

PR #876, opened 2024-11-23
feat: tabs with filters

lukasvinclav:


subham1099:
The PR doesn't work when the tabs have different models (without any filters) - it doesn't highlight the tab correctly when filters are applied.

Without filters: [screenshot](https://github.com/user-attachments/assets/ec57c933-82e7-4f1f-9f2c-17367012668a)
with filters: [screenshot](https://github.com/user-attachments/assets/41e06f9c-c5d8-4540-98ea-9ed14e0b5cb5)

from the code:
```py

            # In case of tabs, we need to check if the query params are the same
            if tabs and query_params != request_params:
                return False
```

shouldn't we compare whether the tabs have query params at all? The current implementation won't highlight for tabs with no query params + manually applied filters

Issue #887, opened 2024-11-29
Move `change_form_before_template` out of form


dickermoshe:
**Do you have any examples?**

No

**Is it a third party application feature request?**

No

**In case of third party application, when it was updated last time?**

No
**Describe your feature request**

The `change_form_before_template` and `change_form_after_template` are so incredibly helpful.
You can add completely custom `html` to the existing views. 

However, right now, they are injected into the `<form>` element of the `changeform` and being hat forms can't be nested, it means that forms can't be included in `change_form_before_template` and `change_form_after_template`

lukasvinclav:
Added in  #891. New properties are `change_form_outer_before_template` and `change_form_outer_after_template`

PR #888, opened 2024-11-29
Move changeform templates outside form tags

dickermoshe:
Move `change_form_before_template` and `change_form_after_template` outside the form tags in `change_form.html`.

Closes: #887

* **Template Changes:**
  - Include `change_form_before_template` outside the form tags.
  - Include `change_form_after_template` outside the form tags.
  - Remove `change_form_before_template` from inside the form tags.
  - Remove `change_form_after_template` from inside the form tags.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/unfoldadmin/django-unfold/pull/888?shareId=d9dcb485-e968-4781-a8d2-f345d3ba27ec).

### DIFF
--- a/src/unfold/templates/admin/change_form.html
+++ b/src/unfold/templates/admin/change_form.html
@@ -61,12 +61,12 @@
     <div id="content-main">
         {% block form_before %}{% endblock %}
 
+        {% if adminform.model_admin.change_form_before_template %}
+            {% include adminform.model_admin.change_form_before_template %}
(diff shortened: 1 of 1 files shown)


lukasvinclav:
We can not move these from the form because there are already special use cases why they need to be inside the form. What we have to do, is to add another two callbacks which are going to display template before and after the template. Now, the hardest part is to find corrent names :-)

PR #896, opened 2024-12-06
docs: cohort correct file path

omargawdat:
the path for the `cohort.html` in the documentation was wrong,  I have just fixed it.


lukasvinclav:
Thanks for contribution!

PR #924, opened 2025-01-02
chore: filters label formatting

einsfr:
Same format as in HTML templates and their {blocktranslate}'s so one msgid may fit them both.


lukasvinclav:
Hey, I just added `lint` workflow to the repository. Would you main to merge the flow from the `main` into your PR? Then we will run the `lint` flow to see if it passes.

lukasvinclav:
Thanks!

PR #954, opened 2025-01-09
feat: date range widget

omargawdat:
## Feature: 
I've implemented support for the PostgreSQL DateTimeRangeField. I kept the implementation straightforward to maintain consistency with the project's existing patterns. Please review and share any feedback you may have.
## Before
image

## After 
image

### DIFF
--- a/src/unfold/admin.py
+++ b/src/unfold/admin.py
@@ -40,6 +40,7 @@ from .widgets import (
     UnfoldAdminImageSmallFieldWidget,
     UnfoldAdminIntegerFieldWidget,
     UnfoldAdminIntegerRangeWidget,
+    UnfoldAdminDateTimeRangeWidget,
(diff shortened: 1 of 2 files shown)


lukasvinclav:
Checked the functionality and it might work for you but we have to choose a different approach. We must use something like `UnfoldAdminSplitDateTimeWidget` (maybe `UnfoldAdminSplitDateTimeRangeWidget`) but instead of two widgets, four separate widgets will be required: date from, time from, date to, and time to. The reason is that we want to display Django admin widgets and browser's native inputs.

I will close the PR now. Not, because it is wrong but because new approach is required. Feel free to create a new PR with logic which I pointed above and I will take a look and I can finish the details (responsive).

omargawdat:
yeah sure will do it

PR #956, opened 2025-01-10
fix(json): properly display JSON strings

Zerotask:
I had a django model with several JSONFields and noticed that one rather big JSON object/string was displayed with double quotes like this:

image

In the database and on non-admin pages where I use a custom template tag, it was displayed properly.

With this little addition, it will be displayed in the admin page properly, too:
image

Other JSONFields will be displayed as before, so this is not a breaking change.

### DIFF
--- a/src/unfold/utils.py
+++ b/src/unfold/utils.py
@@ -141,6 +141,9 @@ def prettify_json(data: Any) -> Optional[str]:
         )
         return highlight(response, JsonLexer(), formatter)
 
+    if isinstance(data, str):
+        data = json.loads(data)
+    
     response = json.dumps(data, sort_keys=True, indent=4)
 
     return mark_safe(
--- a/src/unfold/utils.py
+++ b/src/unfold/utils.py
@@ -143,7 +143,7 @@ def prettify_json(data: Any) -> Optional[str]:
 
     if isinstance(data, str):
         data = json.loads(data)
-    
+
     response = json.dumps(data, sort_keys=True, indent=4)
 
     return mark_safe(


PavelPancocha:
It seems like you are saving "JSON" as a string?

Zerotask:
> It seems like you are saving "JSON" as a string?

Yes, we're saving a JSON string representation, which is a valid value for Django's JSONField.

PavelPancocha:
@Zerotask but then the representation is correct. BTW you are losing the JSONField added value with that - you cannot query the field, etc.

Zerotask:
> @Zerotask but then the representation is correct. 

Not really. Either it should not dump it again, but then you would loose sorting, highlighting and indentation or you just do it like I did and check if it's a string and then load it before. It's just 2 lines and no breaking changes, right.

> BTW you are losing the JSONField added value with that - you cannot query the field, etc.

Yes, that's right. We just use it for logging purposes where we don't want to query it later.

So I get your point, that probably most users would save it as a JSON object not a JSON string, but it's still valid and allowed and this little change here is not worth the discussion, don't you think? The `isinstance` check is really fast and this way both works, JSON objects and JSON strings.

PavelPancocha:
@Zerotask The decision is on @lukasvinclav - I was just walking around and was interested in this. Thanks for discussion. I get you point of view :)

lukasvinclav:
hey, thanks for the discussion guys. Actually, I checked the PR immediately as it was created and your conversation helped me to understand it. I believe that JSONField should be used to store only valid JSON objects. Maybe the specification says something different but if there is a need to store just plain text, use CharField or TextField. In your specific use case, I would just save into DB valid JSON objects (btw logs do not belong into database).

Of course, I'm open to be persuaded that I'm wrong. If somebody has a feedback with updated PR, I will be happy to check it out and change my mind. For now, I'm closing this PR.

PS: The PR contains a bug. If the string is not valid JSON, it will throw an exception so there is needed additional handling.

PR #969, opened 2025-01-15
RTL Support

mohamed-cpp:
I added `rtl:....` to tailwind plugins
that allow us to support RTL language 

Example
Screenshot_1


lukasvinclav:
Thanks for your work. RTL is not going to be a small change and it would require a lot of attention to the details. I believe at least several weeks of the work would be required to polish all the details.

I'm closing this PR, because it is good starter but still, there is no enough resources to put everything together. If there is somebody intereted in this feature, feel free to reach me here:

https://unfoldadmin.com/consulting/

Issue #995, opened 2025-01-22
Adding custom class to widget, removes unfold classes from input


lays147:
**Lib Versions**
python = "^3.10"
django = "^5.1.4"
dynaconf = "^3.2.6"
psycopg2 = "^2.9.10"
django-unfold = "0.45.0"

**What browser are you using?**

Version 133.0.3

**Did you checked changelog/commit history, if the bug is not already fixed?**

Yes

**Did you searched other issues, if the bug is not already fixed?**

Yes

**Did you checked documentation?**

Yes

**Are you able to replicate the bug in the demo site?**

N/A

**Repository with reproduced bug**

TBD

**Describe your issue**

I'm trying to add field masks to some fields in a django model and the custom classes added by unfold are being removed by it. 
```admin.py

class ModelForm(forms.ModelForm):
    product_types = forms.MultipleChoiceField(
        choices=ProductType,
        widget=UnfoldAdminCheckboxSelectMultiple,
        label="Tipos de Produtos",
        required=True,
    )

    class Meta:
        model = Supplier
        fields = "__all__"
        widgets = {"phone": UnfoldAdminTextInputWidget(attrs={"class": "phone-mask"})}

@admin.register(Supplier)
class SuppliferUnfold(ModelAdmin):
    class Media:
        js = ("js/phone-mask.js",)

```
Results in this:

![Image](https://github.com/user-attachments/assets/651d7c3e-73ee-4b73-902e-f628ac3f9a76)
As you can see, all the custom classes are removed and only the "phone-mask" remains:

![Image](https://github.com/user-attachments/assets/d042f7d5-5794-4c68-9c6b-83d0adca3752)

Based on the code of the `UnfoldAminTextInputWidget` the `attrs` with a `class` values overrides the default configuration when this element is present:
```py
class UnfoldAdminTextInputWidget(AdminTextInputWidget):
    def __init__(self, attrs: Optional[Dict[str, Any]] = None) -> None:
        super().__init__(attrs={"class": " ".join(INPUT_CLASSES), **(attrs or {})})
```
As a workaround, I did something like this: 
```py
class UnfoldAdminTextInputWidget(AdminTextInputWidget):
    def __init__(self, attrs: Optional[Dict[str, Any]] = None) -> None:
        classes = " ".join(INPUT_CLASSES)
        if attrs:
            classes += " " + attrs.get("class")
        super().__init__(attrs={"class": classes})
```
[...]

PR #1006, opened 2025-01-26
fix: attrs unpacking overwrite in widgets

lays147:
Hi @lukasvinclav, I'm opening this PR because the issue reported in #995 still persists because the `**(attrs or {})` still overrides the `class` input on the unpacking of the `attrs` object when fetching the `main` branch.

Here you can see that the widgets doesn't have any custom css:
image

Using my fork with this PR I have the css class loaded correctly:
image

Let me know if I need to present other tests.


lukasvinclav:
Thanks for pointing me at the right direction. It was just a wrong order in the code so I fixed it and wrote a proper test #1007

Issue #1011, opened 2025-01-27
feat: title prefix for different environments


lukasvinclav:
The browser tab title should support displaying a custom prefix to enhance the user experience when multiple tabs are open. This will help users easily identify the current site environment.

The functionality should work similarly to the existing ENVIRONMENT setting. To achieve this, a new configuration option, ENVIRONMENT_TITLE, can be introduced.

<img width="454" alt="Image" src="https://github.com/user-attachments/assets/7c622ae8-5f70-4eab-9367-908c0ee92613" />

PR #1015, opened 2025-01-27
feat: title prefix for different environments #1011

mxmaher:
Add a prefix to the browser tab title based on the current configured deployment env, this resolves issue #1011 

**Changes:**
- The environment callback can now return a dict with the `label`, `color_accent` and `title_prefix` keys.
- This change is backward compatible with the current return value as a list of 2 items.
- Added `get_environment_context` to the `UnfoldAdminSite` class that's is called in the `each_context` method to achieve the above changes.
- Adjusted the templates and tests to reflect these changes.

**Screenshots:**
Screenshot from 2025-01-29 05-23-13


mxmaher:
This requires manual testing of the UI which I'm still trying to figure out as I'm facing issues with my devcontainers setup.
Will update the PR with screenshots once I can @lukasvinclav

mxmaher:
@lukasvinclav attached screenshot of the changes in the PR description, looking forward to your feedback

lukasvinclav:
@mxmaher hey, first of thanks for the PR. I was following your changes from the beginning but today I had time to look at it because I wanted to get this functionality in main as soon as possible. 

In last few weeks, I completely refactored chunk of Unfold responsible for settings loading so I was able to create a PR with the same business logic but only on few lines of code. The code which you provided was okay but I wanted even simpler solution. 

In the future, if you are still going to be interested, feel free to push even more PRs. I'm checking all of them, comparing with other solution and waiting for the right moment to merge or reuse them.

PR which I was writing about: #1081

Issue #1027, opened 2025-02-01
Change list filters create URLs with empty key values


dbartenstein:
**What version of Unfold are you using?**

main (version that is running on https://demo.unfoldadmin.com on 1.2.2025 13:00)

**What version of Django are you using?**

What demo.unfoldadmin.com is running

**What browser are you using?**

Firefox 134.0.2 (64-bit)

**Did you checked changelog/commit history, if the bug is not already fixed?**

Yes

**Did you searched other issues, if the bug is not already fixed?**

Yes

**Did you checked documentation?**

No

**Are you able to replicate the bug in the demo site?**

https://demo.unfoldadmin.com/en/admin/formula/driver/

**Repository with reproduced bug**

**Describe your issue**

1. Go to https://demo.unfoldadmin.com/en/admin/formula/driver/
2. Set status filter
![Image](https://github.com/user-attachments/assets/b954270c-0690-4a93-a11f-d041d7ebd568)
3. Click "Apply Filters"

The URL forwarded to contains params like `status__exact` for which no value is set: `https://demo.unfoldadmin.com/en/admin/formula/driver/?salary_from=&status__exact=&created_at_from_1=&created_at_to_1=
`
In this particular case it does not cause harm. But there are many use cases where these params without values lead to errors.

dbartenstein:
In `filterForm`, values are deleted from the original array while looping:
https://github.com/unfoldadmin/django-unfold/compare/0.42.0...0.46.0#diff-d481f8e041ecfcf60429b2635803dba089899b824338dcc7b52f86f27e204349R81

lukasvinclav:
Thanks for report. Fix was already merged into main.

PR #1031, opened 2025-02-03
fix: empty filter query params (#1027)

PavelPancocha:
fix: #1027

### DIFF
--- a/src/unfold/static/unfold/js/app.js
+++ b/src/unfold/static/unfold/js/app.js
@@ -78,9 +78,9 @@ const filterForm = () => {
   }
 
   filterForm.addEventListener("formdata", (event) => {
-    for (const [key, value] of event.formData.entries()) {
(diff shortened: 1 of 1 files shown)


lukasvinclav:
@PavelPancocha what is the actual difference? Does it change something for `if (value === "") event.formData.delete(key);`?

lukasvinclav:
Would you mind to explain what is the purpose of this PR? There is no description to this PR and I'm confused what exactly this PR is supposed to fix.

PavelPancocha:
Linked also the issue to the description. Here is exactly what this addresses: https://github.com/unfoldadmin/django-unfold/issues/1027#issuecomment-2630728405

lukasvinclav:
Okay, finally I was able to understand where I made a mistake. Thanks for the fix. Merging.

Issue #981, opened 2025-01-16
feat: allow customized BooleanFieldListFilter to display with horizontal styling


justinqian42:
**Do you have any examples?**

For example: 

![Image](https://github.com/user-attachments/assets/4feda5d7-e8e1-41e5-a998-270e64de8275)

**Is it a third party application feature request?**

For example: No

**In case of third party application, when it was updated last time?**

For example: N/A

**Describe your feature request**

In the current filter template, line 14 checks if a filter is with the default BooleanFieldListFilter class, and if yes, apply the special horizontal styling.

However, it applies to the default BooleanFieldListFilter only. So when people like me want to customize it to display other options rather than "True" or "False" with a different class name such as "StatusBooleanFieldListFilter", it will display in the default vertical styling.

Suggesting to change it to
`{% if "BooleanFieldListFilter" in spec|class_name %}`
so that as long as people name the inherited filter class with string **BooleanFieldListFilter** in, the styling would work.

Please see the attached screenshot.

Attaching my code snippet (use case) for reference.

`class StatusBooleanFieldListFilter(BooleanFieldListFilter):
    def __init__(self, field, request, params, model, model_admin, field_path):
        self.title = "Status"
        super().__init__(field, request, params, model, model_admin, field_path)

    def choices(self, changelist):
        for lookup, title in ((None, "All"), ("1", "Up"), ("0", "Down")):
            yield {
                "selected": self.lookup_val == lookup,
                "query_string": changelist.get_query_string(
                    {self.lookup_kwarg: lookup}
                ),
                "display": title,
            }`

Please let me know if a PR is preferred. Happy to raise one.

Thanks in advance!

justinqian42:
Raised PR #1033

PR #1033, opened 2025-02-03
feat: style custom BooleanFieldListFilter

justinqian42:


justinqian42:
Ref: #981

lukasvinclav:
I understand what you are trying to achieve but this solution is not generic. What if we want something to display in horizontal mode and at the same time it is not boolean filter? There is another PR which I'm already investigating with more generic approach, adding `horizontal` flag to do class. Anyway thanks for the PR, it helped me to understand the problem.

justinqian42:
Thanks for the explanation @lukasvinclav. Yes, incorporating the flag is a better approach. Thanks for that.

Issue #999, opened 2025-01-23
fix: filtered choices get lost when search used after filtering


sezginacer:
**What version of Unfold are you using?**

main

**What version of Django are you using?**

5.1

**What browser are you using?**

Firefox 135.0

**Did you checked changelog/commit history, if the bug is not already fixed?**

Yes

**Did you searched other issues, if the bug is not already fixed?**

Yes

**Did you checked documentation?**

Yes

**Are you able to replicate the bug in the demo site?**

No, I could not find MultipleDropdownFilter or MultipleRelatedDropdownFilter on demo site.

**Repository with reproduced bug**

No

**Describe your issue**

Suppose we have an admin class like below.

```python
@admin.register(SampleModel)
class SampleModelAdmin(ModelAdmin):
    list_filter = [
        ("status", MultipleDropdownFilter), 
        ("owner", MultipleRelatedDropdownFilter),
        ("writer", MultipleRelatedDropdownFilter),
    ]
```
- Go to change_list page.
- Choose some filters and apply. Then let's say, current URL after filtering is `/admin/app/model/?status=open&status=closed&owner__id__exact=1&owner__id__exact=2&owner__id__exact=3&writer__id__exact=5`
- Write 'test' to search form & click enter. Current URL after search is `/admin/app/model/?q=test&status=open&owner__id__exact=1&writer__id__exact=5`
- So, `status=closed` and `owner__id__exact=2&owner__id__exact=3` filters get lost. 

While just filtering (not searching), there is no problem like this one. But filter a field with multiple values then searching breaks query string & filters.

lukasvinclav:
Would you mind to test again on the most recent version of main?

sezginacer:
When I installed the latest version (0.47.0), nothing has changed, the problem stays the same.

sezginacer:
I have recently opened a [PR](https://github.com/unfoldadmin/django-unfold/pull/1042) for this issue

lukasvinclav:
That PR helped me to debug the issue. Already merged so you can test directly on main. Thanks again

dbartenstein:
@sezginacer unfortunately the merged PR https://github.com/unfoldadmin/django-unfold/pull/1042#issuecomment-2681585926 breaks compatibility with Django 4.2 LTS (supported until April 2026).

I understand the issue you were facing with multiple values for the same parameter.

[...]

PR #1042, opened 2025-02-05
fix: disappearing filters on changelist (#999)

sezginacer:


lukasvinclav:
awesome work. thanks

dbartenstein:
Unfortunately this fix does not work for Django 4.2 LTS, because there’s no `ChangeList.filter_params`.

Even Django main, still uses `cl.params` for compatibility reasons:
https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/search_form.html

@lukasvinclav Unfold admin supports all supported Django versions, I guess? Django 4.2 LTS still has more than one year to live: https://www.djangoproject.com/download/#supported-versions

PR #1056, opened 2025-02-10
chore: font update

lukasvinclav:


edgarsj:
I can confirm these font files work properly.

Screenshot from 2025-02-10 18-52-16

What's the source of these files? I notice they're different from the Inter latest release 4.1  web folder .woff files and around 30kB larger.

lukasvinclav:
@edgarsj you are right! Good eye. I used fonts from different folder (facepalm). Would you mind to test it again with updated fonts from the right folder? Thanks

edgarsj:
I was too lazy to to test properly , so I first checked whether these are the same files I'm already using locally :)

The latest ones are the same, checked them and they work correctly as expected.

PR #1062, opened 2025-02-12
Custom slugify for correct handling of unicode strings

dima23113:
Еdits to process strings specified in Unicode encoding, now tabs will work with Unicode encoding


dima23113:
I figured it out :) you can set the Latin verbose_name directly in NonrelatedTabularInline or another inline and the problem will be solved

PR #1080, opened 2025-02-18
feat: Horizontal layout option for SimpleListFilter using unfold filter template 

andrejbinovsky:


lukasvinclav:
@andrejbinovsky #1092

PR #1097, opened 2025-02-24
fix: tabular inlines total forms

lukasvinclav:


jgeskens (approved):
I tested the change on my project and I can confirm that the error does not occur anymore. Thanks for the quick intervention!

PR #1100, opened 2025-02-24
refactor(qs): remove redundant all() call

Zerotask:
Remove redundant `all()` calls.

References:
- https://docs.djangoproject.com/en/5.1/topics/db/aggregation/#generating-aggregates-over-a-queryset
- https://docs.djangoproject.com/en/5.1/ref/models/querysets/#django.db.models.query.QuerySet.count


Zerotask:
I just saw this when I was looking for something else. There is also potential to reduce the amount of queries:
```python
aggregates = self.q.aggregate(
    min=Min(self.parameter_name),
    max=Max(self.parameter_name),
    total=Count("pk")
)
min_value = aggregates.get("min", 0)
max_value = aggregates.get("max", 0) if aggregates["total"] > 1 else None
```

lukasvinclav:
Would you mind adding a test which will check if the filter is working as expected? For now, I will close the PR but once you have a test, please ping me and I will reopen it.

PR #1108, opened 2025-02-25
fix: filter_params when Django < 5.0

sezginacer:
Fix for filter_params when Django < 5.0. issue [#999](https://github.com/unfoldadmin/django-unfold/issues/999)


src/unfold/compat.py
```diff
@@ -0,0 +1,15 @@
+import django
+from django.contrib.admin.views.main import ERROR_FLAG, PAGE_VAR
+from django.contrib.admin.views.main import ChangeList as DjangoChangeList
+
+if django.VERSION < (5, 0):
```
lukasvinclav, line 5: Do we really need to check for Django version? Isn't it better just to check if there is `filter_params`attribute?

lukasvinclav, line 5: Once we are not checking for Django version, maybe we can move this new class to the views.py exactly as the default Changelist:

https://github.com/django/django/blob/5a1cae3a5675c5733daf5949759476d65aa0e636/django/contrib/admin/views/main.py#L65

src/unfold/compat.py
```diff
@@ -0,0 +1,15 @@
+import django
+from django.contrib.admin.views.main import ERROR_FLAG, PAGE_VAR
+from django.contrib.admin.views.main import ChangeList as DjangoChangeList
```
lukasvinclav, line 3: Rename `DjangoChangeList` to `BaseChangeList`

src/unfold/compat.py
dbartenstein: > Do we really need to check for Django version? Isn't it better just to check if there is `filter_params`attribute?

Checking for the Django version explicitly is a great helper when getting rid of outdated code in the future - as soon as Django 4.2 is deprecated.

Otherwise it will happen that the helper ChangeList will stay there way too long ... just my two cents.

lukasvinclav:
@dbartenstein @sezginacer I did small refactoring where I moved the `ChangeList` inside `views.py` and put the logic inside the `__init__.py`. Please do another test on local and if it works for you, I'm merging to main.

dbartenstein:
> @dbartenstein @sezginacer I did small refactoring where I moved the `ChangeList` inside `views.py` and put the logic inside the `__init__.py`. Please do another test on local and if it works for you, I'm merging to main.

On Django 4.2 LTS it works just fine :clap:

dbartenstein:
@lukasvinclav this bug forces us to revert our upgrade from unfold 0.47 to 0.49.1. Do you plan another unfold release soon? Thanks!

PR #1114, opened 2025-02-28
feat: changelist boolean_text option

jeffbowen:
Adds a ModelAdmin option to show/hide boolean field "True/False/None" text in the changelist. Setting it to `False` allows for narrower boolean field columns like the standard Django admin interface. See "Staff status" and "Active" columns in the examples below:

##### `list_boolean_text = True`
Screenshot 2025-02-28 at 11 04 27 PM

##### `list_boolean_text = False`
Screenshot 2025-02-28 at 11 04 39 PM

##### Standard Django Admin
Screenshot 2025-02-28 at 11 05 06 PM

I prefer the standard, more-compact approach of just showing the icon. Happy to provide screenshots of a changelist with more columns to show how it really helps if needed.

Couldn't find many docs on contributing to this project or its principles, so just let me know if my approach is out of line with what you'd like and I can adjust it.

### DIFF
--- a/docs/configuration/modeladmin.md
+++ b/docs/configuration/modeladmin.md
@@ -45,7 +45,7 @@ class CustomAdminClass(ModelAdmin):
     # Position horizontal scrollbar in changelist at the top
     list_horizontal_scrollbar_top = False
 
-    # Dsable select all action in changelist
+    # Disable select all action in changelist
     list_disable_select_all = False
 
     # Custom actions
--- a/docs/configuration/modeladmin.md
+++ b/docs/configuration/modeladmin.md
@@ -48,6 +48,9 @@ class CustomAdminClass(ModelAdmin):
     # Disable select all action in changelist
     list_disable_select_all = False
 
+    # Set to False to hide text for boolean fields in changelist
+    list_boolean_text = True
+
     # Custom actions
     actions_list = []  # Displayed above the results list
     actions_row = []  # Displayed in a table row in results list
--- a/docs/development/devcontainer.md
+++ b/docs/development/devcontainer.md
@@ -10,7 +10,9 @@ Unfold already contains prepared support for VS Code development. After cloning 
 
 ## Development server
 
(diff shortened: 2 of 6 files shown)


justinqian42:
Good idea. Actually, I prefer the original approach, too.
I doubt if this should be a feature.......since it can be easily customised with a @display decorated function.

jeffbowen:
That's true but if you want *every* boolean field to work this way by default (which I do), my approach seems like the most in line with how other settings are customized in Unfold, right?

In my project that's using Unfold, I'm currently just overriding [`helpers/boolean.html`](https://github.com/unfoldadmin/django-unfold/blob/main/src/unfold/templates/unfold/helpers/boolean.html) like so:

```
{% load i18n %}

{# Overriden boolean template for simpler display #}
<div class="flex items-center ">
    <div class="block rounded-full ml-1 h-2 w-2 {% if value == '' or value == None %}bg-base-500{% elif value %}bg-green-500{% else %}bg-red-500{% endif %}"></div>
</div>
```
Screenshot 2025-03-02 at 8 13 33 PM

It also happens to get rid of the outline-style dot which I like. Was just wanting to find a way to incorporate these options into the project for others. Was going to wait on the outline-style customization until I saw what you guys thought of the `list_boolean_text` option.

Do let me know if there's a better way or if you're not interested in incorporating this into the project.

lukasvinclav:
@jeffbowen thank you very much for the PR. I understand why you need this feature, but for this particular use case I would recommend to just override the template. At the moment, Unfold already contains large amount of features and different options. To keep the project maintainable, I have to be careful what will be merged.

jeffbowen:
Fair enough. Think I mainly just wish the standard Django admin looked a little better. Unfold feels a little heavy for what I'm looking for. Wish there were something in between the current Django admin and Unfold. I don't want to invest a lot of time in customizing Unfold especially once things I'm doing start losing backwards-compatibility with the standard Django admin. Just my two cents FYI. Thanks.

PR #1121, opened 2025-03-04
feat: add styles for django-parler language tabs

metaforx:
Follows the styling of django-modeltranslation and collapsible fieldset. Uses tab instead of pills layout because it's implemented outside of content fieldsets.

django-parler-tabs

django-parler-tabs-mobile


metaforx:
This does apparently not yet support inline tabs and needs some additional work.

Issue #1012, opened 2025-01-27
feat: add a stripe to display the current ENVIRONMENT setting


lukasvinclav:
Enhance the user interface by displaying the current value of the ENVIRONMENT setting as a stripe at the top of the site. This feature will improve environment visibility and can follow the example of [django-admin-env-notice](https://github.com/dizballanze/django-admin-env-notice).

- Use the existing color schemes to style the stripe.
- Extend the environment callback function to include a third argument, which will act as a flag to enable the display of the stripe.
- Update the test suite to cover the new third argument functionality.
- Update the documentation to reflect these changes and include usage examples.

lukasvinclav:
This feature is available in Unfold Studio  and can be tested at https://unfoldadmin.com under "Studio -> Site identity -> Site banner".

PR #1131, opened 2025-03-07
Add option to have a stripe with ENVIRONMENT name on top of the page. Fixes: #1012

SukiCZ:
This adds an option to have a stripe with env name on top of the page

image

The `environment_callback` should be changed to `["Development", "primary", True]`, last argument indicates "show&nbsp;stripe".

Fixes: #1012

### DIFF
--- a/docs/configuration/settings.md
+++ b/docs/configuration/settings.md
@@ -167,10 +167,15 @@ def dashboard_callback(request, context):
 
 def environment_callback(request):
     """
-    Callback has to return a list of two values represeting text value and the color
-    type of the label displayed in top right corner.
+    Callback has to return a list of three values represeting text value and the color
+    type of the label displayed in top right corner. Third value indicates whether the
+    stripe should be displayed on top of the page or not.
     """
-    return ["Production", "danger"] # info, danger, warning, success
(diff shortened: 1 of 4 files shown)


AmineFouzai:
@SukiCZ , You can actually use this, and it will do the same job without major changes. and give the same result 
[django-admin-env-notice](https://github.com/dizballanze/django-admin-env-notice)

and add this env vars as well :
ENVIRONMENT_NAME = "RUNNING ON DEBUG SERVER"
ENVIRONMENT_COLOR = "#FF2222"
ENVIRONMENT_FLOAT = False
ENVIRONMENT_SHOW_TO_UNAUTHENTICATED = False

lukasvinclav:
@AmineFouzai the whole point of this PR is to introduce this functionality into Unfold so there will be no need to install any third party libraries

lukasvinclav:
The catch of this task is to make the environment label fullwidth and not only in the main content area. Probably I would choose little bit different colors. Background need to have `bg-primary-500 dark:bg-primary-500 text-white` for better contrast. If you refactor the code, please ping me on Discord to reopen the PR.

PR #1142, opened 2025-03-11
feat: related objects lookup widget for select/autocomplete fields

CiaranSanders:
Adds the "related objects lookup" functionality from raw id fields for related fields in the admin (either autocomplete or normal select fields). Currently only implemented for Foreign Key/One to One fields; could also look into implementing this functionality for Many to Many fields.

### Examples

Autocomplete field
image

Normal select field
image

Clicking on the magnifying glass will then open the related model's changelist page in a popup. The user can then filter/search for their desired object, and on selection the autocomplete/select field will be populated with the selected choice. 

This is exactly how the raw id field's related object lookup works; this commit just extends that same functionality to other fields in the admin.

I think this is a fantastic feature, but if we think this doesn't belong in Unfold then that is okay! I will continue to use it separately; I thought I might just throw the suggestion out there 😁


CiaranSanders:
Could also add a setting to make this an optional feature, so users can choose to use it or not

src/unfold/templates/unfold/widgets/related_widget_wrapper.html
```diff
@@ -8,6 +8,14 @@
             {% spaceless %}
                 {% if not is_hidden %}
                     {% if can_add_related or can_change_related or can_delete_related or can_view_related%}
+                        {% if can_view_related %}
```
CiaranSanders, line 11: This should probably be something like `if can_edit_field` since there are instances where users can view the related model of the field but are not actually allowed to edit the field

src/unfold/mixins/base_model_admin.py
CiaranSanders, line 1: Utilizing Unfold's `UnfoldRelatedFieldWidgetWrapper` which seemed to have not been in use. But it works perfect for this use case; using it also means we no longer need to set the widgets template manually. The `wrapper_kwargs` setup is copied directly from Django, and is there so we can re-provide the arguments to the related widget wrapper.

lukasvinclav:
This seems like unfinished PR. Please submit a new PR once it is completed.

PR #1154, opened 2025-03-14
feat: list filter buttons on top

CiaranSanders:
Add a setting to display the list filter buttons (submit and clear all filters) at the top of the list filters, instead of at the bottom. This is especially useful for pages with a lot of list filters. Defaults to `False` so that list filters by default still show at the bottom.

### Screenshots

with `list_filter_sheet = False`
image
image

with `list_filter_sheet = True`
image
image


lukasvinclav:
Yeah, I understand. The filters can be sometimes hard to use when you have a lot of fields to filter so sometimes the submit button at the top can be handy. But I think the current implementation is okay and adding new setting will make everything little bit more complex to maintain.

Here, I would choose a different approach. Instead of spending time on this feature, I would focus on making all filters applying immediately after each change. I think this feature can be widely used and it would not require any submit at all ;-)

CiaranSanders:
@lukasvinclav 

The applying all filters immediately on change is nice, until you have the filters in collapsible mode (`list_filter_sheet = True`
...Then its a pain in the butt! 

I know first hand, because I tried that out with my users and they rightfully complained; you have to re-open the filter sheet every time you apply a filter, and that gets really tedious. For when the filters are not collapsed, then applying the filter immediately on change is mostly fine. But even in that case it can be tedious when you know you want to apply multiple filters, and then have to wait each time you click the filter for the page to re-load.

The Apply Filters / Clear Filters buttons at the top were just a minimal solution to that for my pages with a lot of filters. I'll just continue to use this feature personally, but I don't think I will invest time to implement any automatic filtering since I don't find it that useful. Thank you!

PR #1161, opened 2025-03-18
feat(json): allow rendering of custom JSONFields which uses django form JSONField as form_class

Zerotask:
In our case we created a CompressedJSONField which doesn't inherit from models.JSONField but from models.BinaryField. 
In this custom field, we updated the `formfield` method as described here to use `{"form_class": fields.JSONField}`: https://docs.djangoproject.com/en/5.1/howto/custom-model-fields/#specifying-form-field-for-model-field

With this change, it will additionally check for fields.JSONField and it still works with the models.JSONField.
The content will then be rendered as a regular model.JSONField (prettified) instead of just a regular string.

### DIFF
--- a/src/unfold/fields.py
+++ b/src/unfold/fields.py
@@ -9,6 +9,7 @@ from django.db.models import (
     ManyToManyRel,
     OneToOneField,
 )
+from django.forms import fields
 from django.forms.utils import flatatt
 from django.template.defaultfilters import linebreaksbr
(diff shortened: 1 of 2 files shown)


lukasvinclav:
I think this functionality is quite specific for your use case. At the moment Unfold has already plenty of the functionality so it is required to be careful what is going to be merged to keep the possibility of future maintanence realistic.

Issue #943, opened 2025-01-08
feat: hide sidebar globally


vg4tt0:
**Describe your feature request**

Django's default setting `admin.site.enable_nav_sidebar = False` should hide the sidebar collapse icon in the title

lukasvinclav:
What is the expected outcome of this flag? Are we going to completely hide the sidebar if the `enable_nav_sidebar` is set to `False`?

vg4tt0:
That's correct. This is almost the case. Changing this flag to True currently hides the sidebar, however clicking on the icon while on the changeform page still shifts the footer.

lukasvinclav:
Okay, so if the `enable_nav_sidebar` is set to `False`, the sidebar is going to be completely hidden together with the "sidebar toggle" in the header before the title. What about the subpages (e.g. changelist)? Are we going to hide the sidebar there as well?

vg4tt0:
Yes it should be global as originally intended

SukiCZ:
Just to verify, setting `enable_nav_sidebar = False` already hides the sidebar ([Django does that here](https://github.com/django/django/blob/b1324a6/django/contrib/admin/templates/admin/base.html#L84)). The only issue is the collapse icon in the Unfold title, right?

If so, attached PR fixes that.

CleanSlateBot:
@SukiCZ that's correct

vg4tt0:
@SukiCZ that's correct.

SukiCZ:
@vg4tt0 I think it got fixed in #1234 . Can you confirm?

PR #1167, opened 2025-03-19
fix: show dock_to_right only if nav sidebar is enabled

SukiCZ:
When `AdminSite.enable_nav_sidebar` is set to `False`, Django will hide the sidebar ([here](https://github.com/django/django/blob/b1324a6/django/contrib/admin/templates/admin/base.html#L84)) which also applies to Unfold just fine.

The issue is the icon `dock_to_right` is still visible and when user is on a change view, the icon moves the submit buttons (in bottom-right corner) out of the screen.

This PR will show `dock_to_right` only if nav sidebar is enabled.
Fixes: #943


lukasvinclav:
From Django docs:

https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#django.contrib.admin.AdminSite.enable_nav_sidebar

The docs are saying that variable `enable_nav_sidebar` controls if the sidebar is displayed on **larger screens**. In your code you are only controlling display of the toggle button.

SukiCZ:
Yes, control of the nav bar is done [here](https://github.com/unfoldadmin/django-unfold/blob/670d9c46ae20565f15134105359236ff9cab0b72/src/unfold/templates/admin/base.html#L7) already. The only issue IS the icon.

See comment [here](https://github.com/unfoldadmin/django-unfold/issues/943#issuecomment-2738253493) in Issue #943 .

Issue #1019, opened 2025-01-28
Make the top navbar sticky in change_form


arthur-verta:
![Image](https://github.com/user-attachments/assets/7ce5e6dd-4142-4a18-9268-9c232e9f6f5c)

I was thinking to make the top nav bar sticky in changeform to help user navigation.

Are other people interested ?

PR #1168, opened 2025-03-19
feat: make header sticky

SukiCZ:
This will add a sticky header

image

Needs to `npm run tailwind:build`, I didn't to avoid conflicts.
Fixes: #1019


lukasvinclav:
This functionality is okay and it can work if you want to hardcode sticky header in your admin but we need more complex solution which will make the header sticky together with the sidebar header as well. Of course, users should be able to opt-in for sticky header from the Unfold config dict.

PR #1169, opened 2025-03-20
feat: new InfinitePaginator for large tables

lukasvinclav:


Zerotask:
Just a naming suggestion:
I wouldn't call it dumb, because it fulfills a specific purpose.
I'd rather call it `SimplePaginator` or `LightPaginator`

PR #1173, opened 2025-03-21
feat: @display decorator upgrade to show value as link

malefici:
This is how to show a value as a link on the changelist. Yes, I know that button may be used instead, but it is oversized for me.

```python
@admin.register(MyModel)
class TaskAdmin(ModelAdmin):
    list_display = ["link", "new_tab_link"]

    @display(description="Link", link=True)
    def link(self, obj):
        return 'https://example.com', obj.project

    @display(description="New tab link", link=True)
    def new_tab_link(self, obj):
        return 'https://example.com', obj.project, True
```

Sorry, I am not familiar with Python and Django code style is a mystery for me.

### DIFF
--- a/src/unfold/decorators.py
+++ b/src/unfold/decorators.py
@@ -93,6 +93,7 @@ def display(
     dropdown: Optional[bool] = None,
     label: Optional[Union[bool, str, dict[str, str]]] = None,
     header: Optional[bool] = None,
+    link: Optional[bool] = None,
 ) -> Callable:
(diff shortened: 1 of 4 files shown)


lukasvinclav:
I think you can use the code below to render a link. No need to introduce new decorator param.

```python
format_html('<a href="{}">{}</a>', url, text)
```

malefici:
Yes, of course. The main idea is to keep styling and template usage logic. As for button, for example.

PR #1197, opened 2025-04-04
cyrillic support for slugify

Arkhangel:
Django default slugify does not support cyrillic symbols. Added minimal translit for support tabs with cyrillic.


lukasvinclav:
This issue will be fixed here #1229

Arkhangel:
Thanks!

PR #1219, opened 2025-04-11
fix: proper attributions

lukasvinclav:


lukasvinclav:
Yes, you should always run collectstatic when deploying.

PR #1220, opened 2025-04-11
fix: use the same translation for all By {filter_title}

pawel-czarnecki:
This is continuation for https://github.com/unfoldadmin/django-unfold/pull/1146/commits/fb61ed5a8d16ded016a40a185b0a62e16fca7216


lukasvinclav:
Thanks!

PR #1223, opened 2025-04-12
Feat/sticky header

SukiCZ:
Follow up of closed #1168

I've added option to opt-in for sticky header with settings `STICKY_HEADER`

> we need more complex solution which will make the header sticky together with the sidebar header as well.

I don't understand this part of the comment. The header of sidebar already is sticky (always on top when I scroll threw the sidebar content). Could you elaborate please?
image

Needs to `npm run tailwind:build`, I didn't to avoid conflicts.
Fixes: https://github.com/unfoldadmin/django-unfold/issues/1019


SukiCZ:
😢

PR #1229, opened 2025-04-15
fix: inline tabs formset prefix as fragment

lukasvinclav:


artel1992:
So and how it will work? Will we have to set formset for inlines?
Probably can  we add method `slugify` to unfold-config and add custom templatetag where would use it?

artel1992:
Anyway we are waiting it)

devxoul:
@lukasvinclav When will this PR be released? I need this!

PR #1250, opened 2025-04-30
fix: color array support in charts

edgant:
Fix broken support for color arrays introduced in
- feat: css vars support in charts (#945) (36f7dd3)

### DIFF
--- a/src/unfold/static/unfold/js/app.js
+++ b/src/unfold/static/unfold/js/app.js
@@ -289,12 +289,19 @@ const renderCharts = () => {
     for (const key in parsedData.datasets) {
       const dataset = parsedData.datasets[key];
       const processColor = (colorProp) => {
-        if (dataset?.[colorProp]?.startsWith("var(")) {
-          const cssVar = dataset[colorProp].match(/var\((.*?)\)/)[1];
-          const color = getComputedStyle(document.documentElement)
-            .getPropertyValue(cssVar)
-            .trim();
-          dataset[colorProp] = `rgb(${color})`;
+        if (dataset[colorProp]) {
+          const props = Array.isArray(dataset[colorProp])
+            ? dataset[colorProp]
+            : [dataset[colorProp]];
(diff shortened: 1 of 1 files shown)


lukasvinclav:
Do we have an issue where I can read more about this issue? On the demo site, I can change chart colors without any issues.

edgant:
This is useful to to continue supporting when colors are passed as an array of strings instead of string to a chart component:

```python
def dashboard_callback(request, context):
    context.update(
        {
            "kpis": {
                "credits": {
                    "chart": json.dumps(
                        {
                            "labels": [
                                "A", "B", "C"
                            ],
                            "datasets": [
                                {
                                    "data": [1, 2, 3],
                                    "backgroundColor": ["#0EA5E9", "#F43F5E", "#CBD5E1"],
                                },
                            ],
                        },
                    ),
                },
            },
        }
    )
    return context
```

In the template this way:
```html
{% component "components/chart/bar.html" with data=kpis.credits.chart options=kpis.credits.options %}{% endcomponent %}
```

This is especially helpful in doughnut charts (although not officially supported yet) as each dougnut section needs to have different colors:
`components/chart/doughnut.html`
```html
<div class="relative w-full{% if class %} {{ class }}{% endif %}">
  <canvas class="chart" {% if options %}data-options="{{ options }}"{% endif %} data-type="doughnut" data-value="{{ data }}" {% if width %}width="{{ width }}"{% endif %} {% if height %}height="{{ height }}"{% endif %}></canvas>
</div>
```

ECrosland:
Thanks for making this fix. I just encountered the problem it addresses myself when trying to set up a doughnut chart with custom colors.

I'm hoping this can be approved and released soon?

lukasvinclav:
Thanks for the PR. Today, I found a way how to replicate the issue BUT unfortunately the PR was not fixing all edge cases. I created a new PR which should be fine.

https://github.com/unfoldadmin/django-unfold/pull/1592

Issue #1267, opened 2025-05-05
Template tag/variable syntax error in `unfold/layouts/skeleton.html` in 0.56.0


oliverhaas:
**What version of Unfold are you using?**

0.56.0

**What version of Django are you using?**

5.2.0

**What browser are you using?**

None (Django view tests)

**Did you checked changelog/commit history, if the bug is not already fixed?**

Yes. Regression in version 0.56.0 I think; was just released.

**Did you searched other issues, if the bug is not already fixed?**

Yes

**Did you checked documentation?**

Yes

**Are you able to replicate the bug in the demo site?**

No. It's a fairly simple bug so I hope this is enough.

**Repository with reproduced bug**

Just try to render the `unfold/layouts/skeleton.html` template.

**Describe your issue**

Template has a syntax error, specifically using `{{%` where the intent is that the first `{` is an opening bracket of a javascript/alpinejs object, and the `{%` belongs to a template tag, but the template parses interprets the `{{` as a template variable.

See `unfold/layouts/skeleton.html` .

```html
x-data="{{% if opts %}activeTab: 'general',{% endif %} sidebarMobileOpen: false, sidebarDesktopOpen: {% if request.session.toggle_sidebar == False %}false{% else %}true{% endif %} }"
```

From my test suite
```
FAILED anonymous_project/users/tests/test_admin.py::test_view_user - django.template.exceptions.TemplateSyntaxError: Could not parse the remainder: '% if opts %}activeTab: 'general',{% endif %} sidebarMobileOpen: false, sidebarDesktopOpen:...
```

I'll open a PR directly after this.

lukasvinclav:
First of all thanks for the report. I tried to reproduce the issue but I was not to do it for example here:

https://demo.unfoldadmin.com/en/admin/formula/driver/56/change/

Do you know what is the difference? Do you have an exact way how to reproduce the issue so I can avoid it in the future?

Yuliaborzhak:
> So on 5.x it works correctly?

@lukasvinclav 
I don't think so. The author of the issue uses Django 5.2.

lukasvinclav:
@Yuliaborzhak then my question remains. How it is possible that it works on the demo site?

https://demo.unfoldadmin.com/en/admin/

Yuliaborzhak:
> [@Yuliaborzhak](https://github.com/Yuliaborzhak) then my question remains. How it is possible that it works on the demo site?
> 
[...]

PR #1268, opened 2025-05-05
fix: correct template tag/variable syntax in skeleton layout

oliverhaas:
#1267


alorence:
I can confirm the issue is fixed by this PR 👍🏼

PR #1273, opened 2025-05-07
Fix missing pluralization in form error message translation

twFroggen:
Fix missing pluralization in form error message translation

### Issue Description:
When displaying form errors, Unfold's error message ("Please correct the error below.") is not translated correctly, while Django’s built-in admin properly handles localization via pluralization.

This happens because Django admin uses `{% blocktranslate %}` (pluralization), but Unfold uses separate `{% translate %}` calls for singular and plural, causing mismatches with Django's official translation entries.

### Proposed Solution:
Replace separate `{% translate %}` calls with Django’s `{% blocktranslate %}` syntax, matching Django's official admin template.

### Fixes:
- Resolves translation mismatch
- Enables pluralization support consistent with Django’s official translations


lukasvinclav:
Thanks!

PR #1274, opened 2025-05-07
feat: Honor field.encoder in prettify_json

cfra:
When users provide a custom json encoder in their field configuration, this configuration should also be honoured while pretty printing the field values.


lukasvinclav:
Thanks!

PR #1304, opened 2025-05-17
fix: bg in dark mode hover state for hide counts button

### LINKED ISSUE #1303: Filter's button in dark mode uses an inappropriate light hover color
### Description
In dark mode, certain buttons (specifically the "Hide Counts" button in the filter panel) display with a white background color on hover. This creates a visual inconsistency with the dark theme and causes poor contrast with surrounding elements.

https://github.com/user-attachments/assets/70301711-4b20-4084-9747-3bd318aa67c4

### Current Behavior
When hovering over the "Hide Counts" button in dark mode - the button background changes to white

### Expected Behavior
When hovering over buttons in dark mode - the button background should use a dark-appropriate hover color

### Issue

https://github.com/yardensachs/django-unfold/blob/5c1125d219ac747fa61cb81137547290eab2c704/src/unfold/templates/unfold/helpers/change_list_filter_actions.html#L15

The issue is that `hover:bg-base-50` is the only bg class on hover. Its missing `dark:hover:bg-base-900`

yardensachs:
Fixes #1303

### DIFF
--- a/src/unfold/templates/unfold/helpers/change_list_filter_actions.html
+++ b/src/unfold/templates/unfold/helpers/change_list_filter_actions.html
@@ -12,7 +12,7 @@
             <span id="changelist-filter-extra-actions" class="flex flex-row gap-2 items-center mt-2">
                 {% if cl.is_facets_optional %}
                     {% if cl.add_facets %}
-                        <a href="{{ cl.remove_facet_link }}" class="hidelink border grow font-medium px-3 py-2 rounded-default text-center transition-all w-full hover:bg-base-50 lg:w-auto dark:border-base-700 dark:hover:text-base-200">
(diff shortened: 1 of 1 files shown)


lukasvinclav:
Thanks for the PR. I just checked the functionality on the demo site and it works correctly for me:

http://localhost:8000/en/admin/formula/driverwithfilters/

Would you mind testing it? Please refresh the local cache before test. Thanks

yardensachs:
@lukasvinclav I was trying this one:
https://demo.unfoldadmin.com/en/admin/formula/driverwithfilters/?_facets=True

You need to click on show counts first, and its only an issue with Hide Counts button

lukasvinclav:
Thanks for the PR. I checked the code and I noticed the trick was removing wrong background color:

#1314

PR #1331, opened 2025-05-24
feat: automatically convert standard model admins for unfold

gearheart:
When model admin is not inherited from unfold's ModelAdmin - then the form looks unstyled. It's easy to avoid with own applications, but makes it a little tedious to install and use third-party apps - because I need to unregister and replace builtin admin classes.
I thought it would be great if third party apps would just work at least for simple cases. So I made this simple slightly magic trick, that does work for me at least on simple cases.

When new "unfold.contrib.convert" is added to INSTALLED_APPS (in the last position) it:
    1. Finds all admin classes that are not subclasses of unfold.admin.ModelAdmin
    2. Unregisters them
    3. Creates new classes that inherit from both the original admin and unfold.admin.ModelAdmin
    4. Fixes any inlines by creating new classes that inherit from unfold inline classes
    5. Registers the new classes

This is a very basic conversion, but it does reduces a friction of using unfold with third party apps for me.

### DIFF
--- a/src/unfold/contrib/convert/apps.py
+++ b/src/unfold/contrib/convert/apps.py
@@ -1,0 +1,16 @@
+from typing import override
+
+from django.apps import AppConfig
+
+
+class DefaultAppConfig(AppConfig):
+    name = "unfold.contrib.convert"
+    label = "unfold_convert"
+
+    @override
+    def ready(self) -> None:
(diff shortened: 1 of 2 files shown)


diesieben07:
It would be nice if this could have a list of extended conversions for built-in integrations like django-import-export that require more changes, such as changing the form.

PR #1376, opened 2025-06-17
fix simple_history buttons

### LINKED ISSUE #1375: Simple History revert and change buttons do not work

0.59

5.2

Edge 137.0.3296.83

Yes

Yes

Yes

"Revert" and "Change History" buttons are not shown when logged into the demo user.

**Repository with reproduced bug**

Add the following code to formula.admin.DriverAdmin:

```python
class DriverAdmin(GuardedModelAdmin, SimpleHistoryAdmin, DriverAdminMixin):
    ...
    def revert_disabled(self, request, obj=None):
        return True
```

The revert button should show now and you can test it out for yourself.

**Describe your issue**

The revert and change history buttons do not work. The fix is quite simple, so I submitted a pull requests and will link back to it.

JBrut22:
The "Revert" and "Change History" buttons in historical records are not able to be clicked. This is how I modified unfold/contrib/simple_history/templates/submit_line.html to fix it on my end. It is a relatively simple fix and I figured others could benefit from it.

I also added the _cursor-pointer_ css class so that it indicates to the user it can be pressed when hovering over it.

Closes #1375


lukasvinclav:
Thanks!

PR #1382, opened 2025-06-19
feat: added icon for collapsible inlines

CiaranSanders:
Adds the same icon used for collapsible rows in changelist to the collapsible inlines

Before:
image

After:
image

Icon is pulled from the changlist dropdown:
image

Minimal change which I think helps keep everything feel consistent. Plus, it was a little unintuitive that the inlines were dropdowns for my users, the icon helps indicate so to them.


lukasvinclav:
Hey, thanks for the PR. I just added there quick fixes for:

- icon for collapsible inlines
- wrong alignment when sorting is used at the same time

ciaransanders:
Ah whoops, thanks for fixing that 😁

PR #1383, opened 2025-06-20
fix: add missing template override for django-import-export

### LINKED ISSUE #1372: Missing template override for django-import-export

0.59.0

For example: 5.1.5

Chrome 137

Yes

Yes

Yes

No

**Repository with reproduced bug**

n/a

**Describe your issue**

When using django-import-export to create an Admin which supports _only_ importing (`class MyAdmin(ImportMixin, ModelAdmin)`), django-import-export will use the template `admin/import_export/change_list_import.html`. This template is not overridden by unfold, causing the "Import" button to appear in the wrong place:

Image

This can be fixed easily by adding the missing template override, this has already been done for `change_list_export.html` (which is used if only exporting is enabled).

diesieben07:
Fixes #1372

### DIFF
--- a/src/unfold/contrib/import_export/templates/admin/import_export/change_list_import.html
+++ b/src/unfold/contrib/import_export/templates/admin/import_export/change_list_import.html
@@ -1,0 +1,5 @@
+{% extends "admin/import_export/change_list.html" %}
+
+{% block actions-items %}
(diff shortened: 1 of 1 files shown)


lukasvinclav:
Awesome. Thanks for contribution!

PR #1403, opened 2025-06-24
feat: warning icon for form errors on tabs

ciaransanders:
This feature just simply introduces a small styling change for fieldset and inline tabs, so when a form error occurs, every fieldset and inline tab containing a field with an error now indicates it has an error.

Currently, when a form error occurs and fieldset tabs are in use, it is not clear in which tab the error occured under:
image

Example here a field under the "Other Information" tab had an error:
image

An example when multiple tabs have fields with form errors:
image

Example when multiple inlines with tabs had errors:
image


ciaransanders:
While testing this some more, I was getting weird behaviors in some model admins where the `is_valid` method was evaluating to False when it should not have been... Might be just a misunderstanding on my part on how that function is supposed to work.

But I replaced it's usage with `total_error_count` for determining if an inline tab has any errors under it.

lukasvinclav:
@ciaransanders fieldset tabs are now supported https://github.com/unfoldadmin/django-unfold/pull/1779

lukasvinclav:
@ciaransanders inline tabs are now supported :-) https://github.com/unfoldadmin/django-unfold/pull/1799

ciaransanders:
Awesome work, thanks @lukasvinclav !

PR #1430, opened 2025-07-02
feat: list_fullwidth for custom pages

khadeom:
Added list_fullwith feature for custom pages similar to test_fullwith in ModelAdmin

### DIFF
--- a/src/unfold/templates/admin/base.html
+++ b/src/unfold/templates/admin/base.html
@@ -32,14 +32,14 @@
 
             {% block messages %}
                 <div class="px-4">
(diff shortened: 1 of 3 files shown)


khadeom:
@lukasvinclav Could you please share your thoughts on this implementation?
I’m wondering if we can avoid the double check in base.html:
```django
{% if not cl.model_admin.list_fullwidth and not list_fullwidth %}
```

Is there a cleaner way to handle this condition—perhaps a unified flag or context variable that works for both the changelist and custom pages?

PR #1463, opened 2025-07-19
feat: sortable changelist

lukasvinclav:


thelittlebug:
i tried that feature, and it works for me.
but maybe there needs to be some way to move the things across the pagination boundaries. take a look at django sortable. a good solution? no. but its working

lukasvinclav:
@thelittlebug what do you mean by "to move the things across the pagination boundaries"?

PR #1464, opened 2025-07-19
fix: trailing action url slash

lukasvinclav:


alfonsrv:
ah sorry, didn't see you already fixed it
can close this then https://github.com/unfoldadmin/django-unfold/pull/1467

PR #1485, opened 2025-08-01
fix: removed the redundant "All" option from MultipleDropdownFilter

yardensachs:
This fixes an issue where multi-value dropdown has an "all" option, which does not do anything and is redundant:
SCR-20250731-snqf
SCR-20250731-snnd

This was solved for checkboxes, so I am just following the pattern in checkboxes:
https://github.com/unfoldadmin/django-unfold/blob/bbc6da46fd36504e233d4d835908632785081f30/src/unfold/contrib/filters/admin/choice_filters.py#L27-L31


lukasvinclav:
hey, thanks for the PR. I already tested your code and found other dropdowns are having the same issue. I will try to write a PR all edge cases.

yardensachs:
@lukasvinclav Thanks! is there another PR I can follow for this?

PR #1494, opened 2025-08-04
Improve translation for "delete" in admin.

FranzDiebold:
(i.e. better for german translation etc.)


lukasvinclav:
hey Franz, I understand what you are trying to improve but I want to keep a compatibility with Django admin translations:

https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/submit_line.html#L14

When we are talking about DE translations. Do you think, you can help me to translate remaining strings here?

https://github.com/unfoldadmin/django-unfold/pull/1477

PR #1498, opened 2025-08-07
fix: disable sort functionality on non-sortable changelist

ilhamelhamdi:
Hi Lukas!

I've tried sortable changelist feature and works quite well.
But I've seen a bug for non-sortable changelists. They are draggable even though they shouldn't be.

prev_sortable_changelist

I tried to fix this issue in this PR.
I hope this feature can be officially released soon. Thank you


lukasvinclav:
thank for PR. Wrap `x-on:end="sortRecords"` into condition as well.

ilhamelhamdi:
> thank for PR. Wrap `x-on:end="sortRecords"` into condition as well.

Alright then. Thanks for the review, Lukas!

Issue #1484, opened 2025-07-31
Conditional Fields Empty on Hidden


lucmerceron:
**What version of Unfold are you using?**

main

**What version of Django are you using?**

5.2.2

**What browser are you using?**

All

**Did you checked changelog/commit history, if the bug is not already fixed?**

No

**Did you searched other issues, if the bug is not already fixed?**

Yes

**Did you checked documentation?**

Yes

**Are you able to replicate the bug in the demo site?**

Yes

**Repository with reproduced bug**

N/A

**Describe your issue**

### Steps to reproduce:
1. Go on live demo on this page: https://demo.unfoldadmin.com/en/admin/formula/driver/56/change/
2. Status: Active
3. Fill the `Conditional field active` with "Fake active data"
4. Change status to Inactive
5. Fill the `Conditional field inactive` with "Fake inactive data"
6. Submit

### Current behavior:
The `Conditional field active` is also sent with the payload
```
...
------geckoformboundary5bdd477ff4cc9b7773a4f4a54b98733c
Content-Disposition: form-data; name="status"

ACTIVE
------geckoformboundary5bdd477ff4cc9b7773a4f4a54b98733c
Content-Disposition: form-data; name="conditional_field_active"

Fake active data
------geckoformboundary5bdd477ff4cc9b7773a4f4a54b98733c
Content-Disposition: form-data; name="conditional_field_inactive"

Fake inactive data
...
```

### Expected behavior:
We would expect Unfold to clear / empty the hidden fields before submitting

```
...
------geckoformboundary5bdd477ff4cc9b7773a4f4a54b98733c
Content-Disposition: form-data; name="status"

ACTIVE
------geckoformboundary5bdd477ff4cc9b7773a4f4a54b98733c
Content-Disposition: form-data; name="conditional_field_inactive"

Fake inactive data
...
```

lukasvinclav:
Hey, thanks for feedback. At the moment, conditional fields are only responsible for displaying/hiding form inputs via CSS(`x-show` in alpinejs) and there is no other business logic related to form processing on the backend. If you need some additional form handling then I suggest overriding the model form (`form = SomeForm`) and changing form fields accordingly.

lucmerceron:
Hello @lukasvinclav,
Thanks for your response!

Conditional fields are often used to represent specific constraints enforced by the API such as mutually exclusive combinations of fields. Ideally, these constraints would raise validation errors if violated.

[...]

PR #1540, opened 2025-08-26
fix(conditional): disabled on change

lucmerceron:
Following discussion on https://github.com/unfoldadmin/django-unfold/issues/1484


lucmerceron:
Hello @lukasvinclav, I spent some time on that PR, could you please tell me what was wrong with it?

lukasvinclav:
@lucmerceron sorry for closing the ticket without the further explanation. At the moment, conditional fields are pretty complex and just investigation and writing answer to simple questions takes a lot of time.

To be honest, I'm not sure what we want to achieve here. What is supposed to be end result? I read your issue and I think you want prevent sending the data from the fields which are conditionally NOT displayed. From my perspective, maybe I already mentioned it somewhere, this should be not handled by FE because on BE you are going to write your custom validation logic.

lukasvinclav:
@lucmerceron if there will be any feedback, let me know. For now, I will close the PR.

Issue #1545, opened 2025-08-31
Header title doesn't support translated strings


Yuliaborzhak:
**What version of Unfold are you using?**
0.65.0

**What version of Django are you using?**
4.2

**What browser are you using?**
Not related

**Did you checked changelog/commit history, if the bug is not already fixed?**
Yes

**Did you searched other issues, if the bug is not already fixed?**
Yes

**Did you checked documentation?**
Yes

**Are you able to replicate the bug in the demo site?**
No. Does it have custom pages?

**Repository with reproduced bug**
N/A

**Describe your issue**
On custom pages a template tag `header_title`  breaks on this line with `AttributeError: '__proxy__' object has no attribute '_meta'`:
https://github.com/unfoldadmin/django-unfold/blob/main/src/unfold/templatetags/unfold.py#L628.

Reason:
We use gettext_lazy for context["original"] strings. 
```
context["original"] = _("Some value")
```
It doesn't return a string, it returns StrOrPromise. As a result, a condition: `not isinstance(original, str)` passes and the code goes into a wrong if-branch.

Actual behavior:
context["original"] = _("Some value") execution goes to 
https://github.com/unfoldadmin/django-unfold/blob/main/src/unfold/templatetags/unfold.py#L628

Expected behavior:
context["original"] = _("Some value") execution  goes to 
https://github.com/unfoldadmin/django-unfold/blob/main/src/unfold/templatetags/unfold.py#L666

I have created a PR to fix it.

lukasvinclav:
My understanding is that you are setting `context["original"] = _("Some string")`, but in Django admin changeform `original` value is reserved for current object (model instance). Are you trying somehow overriding the default `__str__` return value of model instance?

Yuliaborzhak:
@lukasvinclav 
No, we have a different use case. We ran onto the problem on confirmation pages that perform bulk actions, so actually there is no current _object_ as you deal with multiple _objects_ at once.  So we are overriding directly in a confirmation page view  as follows: `context["original"] = _("Bulk action name")` .

lukasvinclav:
is It possible to replicate this on demo site? I tried to do a bulk delete from changelist but on confirmation screen I don't see any errors

Yuliaborzhak:
@lukasvinclav 
[...]

PR #1546, opened 2025-08-31
fix: header title for translated strings

Yuliaborzhak:
See #1545


src/unfold/templatetags/unfold.py
```diff
@@ -625,7 +626,9 @@ def header_title(context: RequestContext) -> str:
             }
         )
 
-        if (original := context.get("original")) and not isinstance(original, str):
+        if (original := context.get("original")) and not isinstance(
+            original, StrOrPromise
```
PavelPancocha, line 630:
```
from django.utils.functional import Promise

is_text = isinstance(original, (str, Promise))
```

PR #1552, opened 2025-09-03
Add null check for placeholder before setting value

Mte90:
I am using django-file-form because I need an ajax upload but this js code generate an error in the browser as the HTML is different.

### DIFF
--- a/src/unfold/static/unfold/js/app.js
+++ b/src/unfold/static/unfold/js/app.js
@@ -352,7 +352,9 @@ const fileInputUpdatePath = () => {
           input.parentNode.parentNode.parentNode.querySelector(
             "input[type=text]"
           );
-        placeholder.setAttribute("value", parts[parts.length - 1]);
+        if (placeholder !== null) {
+          placeholder.setAttribute("value", parts[parts.length - 1]);
+        }
       });
 
       input.hasChangeListener = true;


Mte90:
I got the same issue with a custom widget that uploads video in the admin to the same line.

Mte90:
To help test it, this happens with custom widget that don't have a placeholder above so it generates an error in js that block everything in the page itself.

Mte90:
why it was closed?

Mte90:
I have a widget with this HTML:

```
<div class="bg-white flex flex-none items-center leading-none self-stretch dark:bg-base-900">
    <input type="file" id="{{ widget.attrs.id }}" accept="video/*" class="border border-base-200 bg-white font-medium min-w-20 placeholder-base-400 rounded-default shadow-xs text-font-default-light text-sm focus:outline-2 focus:-outline-offset-2 focus:outline-primary-600 group-[.errors]:border-red-600 focus:group-[.errors]:outline-red-600 dark:bg-base-900 dark:border-base-700 dark:text-font-default-dark dark:group-[.errors]:border-red-500 dark:focus:group-[.errors]:outline-red-500 dark:scheme-dark group-[.primary]:border-transparent disabled:!bg-base-50 dark:disabled:!bg-base-800 px-3 py-2 w-full max-w-2xl " />
</div>
```

Without that fix I am not able to use it

PR #1556, opened 2025-09-06
fix #1528 django-constance: can't disable boolean setting

### LINKED ISSUE #1528: django-constance: can't disable boolean setting when using unfold's additional choice field

0.64.2

5.1

Latest Safari

Yes

Yes

Yes

Yes, see "Website in construction" and try to deactivate it

https://demo.unfoldadmin.com/de/admin/constance/config/

**Describe your issue**

when using unfold'd choice field for a boolean setting, you can't deactivate it (set it to false). This will raise a validation error.

This is the default:

Image

Now I uncheck and save it:

Image

Image

If I reload the page, it's still checked (True).

When I remove `CONSTANCE_ADDITIONAL_FIELDS` and use constance's default, I can uncheck it

titusz:
The issue occurs because Django Unfold's constance integration uses django.forms.BooleanField with the default
  configuration, which has required=True by default. In Django's BooleanField:
  - required=True means the checkbox MUST be checked for the form to validate
  - required=False means the checkbox can be either checked (True) or unchecked (False/None)

  When a user tries to uncheck a boolean setting in django-constance through Unfold's interface, the form validation
   fails because the BooleanField expects the checkbox to be checked (since required=True).

### DIFF
--- a/src/unfold/contrib/constance/settings.py
+++ b/src/unfold/contrib/constance/settings.py
@@ -15,6 +15,7 @@ UNFOLD_CONSTANCE_ADDITIONAL_FIELDS = {
         "django.forms.BooleanField",
         {
             "widget": "unfold.widgets.UnfoldBooleanSwitchWidget",
+            "required": False,
         },
     ],
     "file_field": [


lukasvinclav:
Thanks for your contribution!

PR #1563, opened 2025-09-08
feat: added styling to highlighted items in select2 dropdown

ciaransanders:
Adds styling to highlight the currently selected item in the dropdown of autocomplete widgets when using the arrow keys (up and down) to navigate in the dropdown.

In these examples, "Pump" was already selected in the autocomplete widget, and "S&T Assembly" was navigated to using the arrow keys:
image
image

Also works with multiple choice autocomplete widgets.


lukasvinclav:
Thanks!

PR #1567, opened 2025-09-12
feat: add template option to display decorator

ohrstrom:
Adds a `template` option to `@display` decorator.

The `template` option expects a Django template, which is rendered with the provided context.

```python
class DemoAdmin(ModelAdmin):
    list_display = [
        "display_template",
    ]

    @display(
        description="Template",
        template="demo/admin/demo_template.html",
    )
    def display_template(self, obj):
        return {
            "name": obj.name,
            "count": obj.related.count(),
        }
``` 

```html
<!-- demo/admin/demo_template.html -->
{{ name }} <sup>{{ count }}</sup>
```

This makes it easier to render arbitrary templates directly through the decorator, without needing to call `render_to_string` in the admin classes.

Would this be helpful?

### DIFF
--- a/src/unfold/decorators.py
+++ b/src/unfold/decorators.py
@@ -106,6 +106,7 @@ def display(
     dropdown: Optional[bool] = None,
     label: Optional[Union[bool, str, dict[str, str]]] = None,
     header: Optional[bool] = None,
+    template: Optional[str] = None,
 ) -> Callable:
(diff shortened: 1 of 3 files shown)


lukasvinclav:
Hey, first of all thanks for the idea. On one side, I understand why you need that but from my perspective I don't want push more logic into actions at the moment. Your implementation is fine but it still requires docs, tests and some examples in the demo project.

PR #1575, opened 2025-09-17
Init Czech translations

SukiCZ:
Initialize (Czech :czech_republic: ) translations


SukiCZ:
After internal discussion, lukasvinclav will do infrastructure for translation first.

PR #1570, opened 2025-09-14
feat!: nested inlines

lukasvinclav:

PR #1619, opened 2025-10-14
Fix inline add permission check in tabular template and JavaScript

twFroggen:
- Template: Only set id='empty' when has_add_permission is True
- JavaScript: Check template existence before cloning (click handler)
- JavaScript: Check has_add_permission before showing add button
- Align with Django's original tabular.html template behavior
- Prevents users from adding inline items when has_add_permission returns False

Fixes issue where add button was clickable even without permission, causing JavaScript errors when attempting to clone non-existent template.


lukasvinclav:
Thanks for the PR. There is new functionality regarding the inlines which is fixing this issue.

https://github.com/unfoldadmin/django-unfold/pull/1570

Issue #1645, opened 2025-10-28
RangeDateTimeFilter does not work with localized formats


brandenhall:
**What version of Unfold are you using?**
0.69.0

**What version of Django are you using?**
5.2.7

**What browser are you using?**
Chrome 141.0.7390.123

**Did you checked changelog/commit history, if the bug is not already fixed?**
Yes

**Did you searched other issues, if the bug is not already fixed?**
Yes

**Did you checked documentation?**
Yes

**Are you able to replicate the bug in the demo site?**
No

**Repository with reproduced bug**
https://github.com/brandenhall/django-unfold-datetime-formatting-issue

**Describe your issue**
When using a localized formats the RangeDateTimeFilter does not work. 
The linked repository has a localized format for English and is installed via the Django `FORMAT_MODULE_PATH` setting.
It also has a sample database an Event model. There are 100 instances that have `created` datetime fields. There is a RangeDateTimeFilter setup to filter these instances and it does not work and appears to fail silently.

PR #1646, opened 2025-10-28
Fix: parse_date_str and parse_datetime_str respect localized formats

brandenhall:
This is a fix for [issue 1645](https://github.com/unfoldadmin/django-unfold/issues/1645)

Rather than directly pulling `DATE_INPUT_FORMATS` and `DATETIME_INPUT_FORMATS` directly from `settings`, this fix now gets those values via a call to `formats.get_format`, which handles getting the correct values from Django's localization system.

### DIFF
--- a/src/unfold/utils.py
+++ b/src/unfold/utils.py
@@ -4,7 +4,6 @@ import json
 from collections.abc import Iterable
 from typing import Any
 
-from django.conf import settings
 from django.db import models
 from django.db.models import Model
 from django.template.loader import render_to_string


lukasvinclav:
@brandenhall I was testing the implementation but I think it does not support custom formats from `settings.py`. Example:

```python

# settings.py
DATE_INPUT_FORMATS = [
    "%d.%m.%Y",  # Custom input
    "%Y-%m-%d",  # '2006-10-25'
]
```

lukasvinclav:
@brandenhall any update?

brandenhall:
Yeah, I investigated this further and you're correct - this is a bit embarrassing, but also a quirk of Django. I combed through the Django source code and input formats are ONLY handled by settings.py. The localized formats.py is only for output formats, not input. I'm closing this PR. Thank you for checking!

PR #1671, opened 2025-11-07
fix: use minifed version of jquery when settings.DEBUG is disabled

zaggino:
Similar approach as can be seen in `django/contrib/admin/widgets.py`, otherwise unfold causes two jQuery versions to be loaded on a single page (both regular and minified version) and that breaks select2 controls with an error:

image

Ref: https://github.com/django/django/blob/main/django/contrib/admin/widgets.py#L594-L602

### DIFF
--- a/src/unfold/widgets.py
+++ b/src/unfold/widgets.py
@@ -1,5 +1,6 @@
 from collections.abc import Callable
 from typing import Any
+
 from django.conf import settings
 from django.contrib.admin.options import VERTICAL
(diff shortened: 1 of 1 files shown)


lukasvinclav:
@zaggino awesome catch. Would you mind fixing `widgets.py` as well? Once fixed, I will merge it asap. Thanks!

zaggino:
@lukasvinclav sure, done... updated the `action-form-example.md` as well

lukasvinclav:
thanks!

PR #1688, opened 2025-11-14
style: no highlighting for user link icons

ciaransanders:
Add `select-none` tailwindcss class to user link icons in top right, just so they stop highlighting whenever I try to click them and accidentally drag my mouse (which is all the time).


lukasvinclav:
hey, nice touch but these icons will stop to exists after merging this PR https://github.com/unfoldadmin/django-unfold/pull/1684. Top right navigation, will be moved at the bottom of sidebar.

PR #1689, opened 2025-11-16
fix: remove `truncate` class to restore dropdown scrollbar visibility

ali-hv:
### Summary
This PR removes the `truncate` class from the default SELECT_CLASSES configuration.   Using `truncate` on a native `<select>` applies `overflow: hidden` and `white-space: nowrap`, which prevents the browser from rendering the scroll bar inside the dropdown list.

### What Was Happening
- When the select menu contained many options, the dropdown would open but **no scrollbar was shown**.
- This behavior came from the `truncate` utility (Tailwind), which is not compatible with native `<select>` elements.

### Fix
- Removed the `truncate` class from SELECT_CLASSES so the dropdown can render normally with scroll support.

### Impact
- Scroll bar now appears correctly when a select widget contains many items.
- No visual regressions in the collapsed select display.

### DIFF
--- a/src/unfold/widgets.py
+++ b/src/unfold/widgets.py
@@ -130,7 +130,6 @@ SELECT_CLASSES = [
     "pr-8!",
     "max-w-2xl",
     "appearance-none",
-    "truncate",
 ]
 
 PROSE_CLASSES = [


lukasvinclav:
Would you mind creating an issue first and answering all question there? I'm not sure what exactly is broken in current dropdown.

ali-hv:
> Would you mind creating an issue first and answering all question there? I'm not sure what exactly is broken in current dropdown.

I did:
https://github.com/unfoldadmin/django-unfold/issues/1690

lukasvinclav:
Thanks for PR. But just removing `truncate` will break other edge cases. I fixed the issue in different PR https://github.com/unfoldadmin/django-unfold/pull/1702

PR #1698, opened 2025-11-19
feat: sidebar badge styles

lukasvinclav:


Elendiar:
@lukasvinclav btw, it was not necessary to make badge text uppercase.

lukasvinclav:
@Elendiar please, create a new issue where you will describe your use case so I can understand how exactly you want to use badges. thanks

alfonsrv:
What if badges are empty / have no content? Their container is still rendered.

PR #1704, opened 2025-11-21
Add Italian translation

Alexclne:
This PR adds the full Italian translation for django-unfold.
All strings in django.po were translated and compiled.


lukasvinclav:
Thank you very much for the PR. Before we can merge this PR, we need to find out and prepare the documentation how to extract strings for translations from the project. It is important that these strings are not going to contain already translated strings from the django admin so we will not translate something what is already completed.

I'm going to close this ticket for now, but once I will start working on this topic, I will check what I can do here.

PR #1720, opened 2025-12-03
fix: Actually paginate autocomplete JSON response

dfr-exnaton:
Currently the response of the autocomplete view returns always the whole queryset.


lukasvinclav:
Hey, thank you very much for PR. I was able to reproduce the error and additionally found one more thing to fix. I updated the docs as well. 

https://github.com/unfoldadmin/django-unfold/pull/1783/files

PR #1820, opened 2026-01-28
fix: Show calendarbox above background

ghostishev:
Problem: After raising `changelist-filter` z-index in https://github.com/unfoldadmin/django-unfold/pull/1773/changes , calendarbox z-index remain unchanged and date picker appears under blurred background

Solution: Raise z-index for `.calendarbox .calendar` to 60 as well.

Validation:


lukasvinclav:
Thanks for report. Fixed here https://github.com/unfoldadmin/django-unfold/pull/1837

ghostishev:
Thank you!

PR #1825, opened 2026-01-29
fix: add hovers to primary buttons

Yuliaborzhak:
**Normal state:**
image

**Hover state:**
image

**Normal state:**
image

**Hover state:**
image

Css needs to be regenrated for `hover:border-primary-700` to be added.

I wasn't addressing a Button conponent in this PR. All Button variants need hovers reworked. I think it makes more sense to do them all at once to keep the component consistent, which is out of scope of this PR.
https://github.com/unfoldadmin/django-unfold/blob/9d048e1776f10e1496617b9a0caf930b10d8541c/src/unfold/templates/unfold/components/button.html#L10

### DIFF
--- a/src/unfold/templates/admin/actions.html
+++ b/src/unfold/templates/admin/actions.html
@@ -20,7 +20,7 @@
                 {% endblock %}
 
                 {% block actions-submit %}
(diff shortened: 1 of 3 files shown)


lukasvinclav:
This is already in my TODO but I will solve it in different way. All buttons will be replaced by button component containing new hover classes.

Issue #1821, opened 2026-01-28
End2end testing support


Yuliaborzhak:
Unfold works great with unit testing. What about adding an end2end testing support? Currently most of html elements in Unfold don't have individual identifiers (ids, classes, data- attributes whatever) so locating elements requires something like:
`self.page.locator('#nav-sidebar div[x-show="openDropdown"] a:last-child')`
<img width="1884" height="764" alt="Image" src="https://github.com/user-attachments/assets/3eb5c746-32c4-4969-a4a6-2f9b7438a161" />
which is not really stable and semantic.

Suggested solutions:
1. Adding data-testids on main structural elements (playwright support).
https://playwright.dev/python/docs/api/class-page#page-get-by-test-id
OR
2. Adding semantic css classes on main structural elements (generic support). 

Thank you for considering it!

lukasvinclav:
If you need a specific `data-testid`, just create a PR with explanation. If it is going to make sense, I will merge it.

Yuliaborzhak:
@lukasvinclav Thank you. Done.
https://github.com/unfoldadmin/django-unfold/pull/1826

PR #1826, opened 2026-01-29
feature: add data_testids

Yuliaborzhak:
See 
- https://github.com/unfoldadmin/django-unfold/issues/1821


lukasvinclav:
Feedback:

- I think most of these can be replaced by id, instead of `data-testid` so it can be reused for styling purposes as well.
- Adding attributes for testing purposes without including a test?
- You can target last element without a specific class. CSS has a selector for that.
- Names can be improved.
- Use "-" instead of "_".

I'm going to close this one because there is too much work required to finish this PR. If I were you, I would split it into several smaller ones and go through them one by one. The smaller the PR, the higher the chance I will approve it.

PR #1877, opened 2026-02-17
fix: tabs_primary_active crashes with 'str has no attribute opts' on non-changeform views

laurentquillerou:
When `inline_admin_formsets` is not in the template context on changelist/custom pages, django resolves it as `""`

It was introduced in 0.77.0

### DIFF
--- a/src/unfold/templatetags/unfold.py
+++ b/src/unfold/templatetags/unfold.py
@@ -896,6 +896,9 @@ def tabs_errors_count(fieldset: Fieldset) -> int:
 def tabs_primary_active(inlines: list[InlineAdminFormSet]) -> str:
     active = "general"
 
+    if not isinstance(inlines, list):
+        return active
+
     for inline in inlines:
         if getattr(inline.opts, "tab", False) and inline.formset.errors:
             for error in inline.formset.errors:


lukasvinclav:
Please create a ticket where you will describe how I can reproduce the issue. Right now, I have no clue what we are fixing.

laurentquillerou:
In the app `models.py`, we have models like:
```py
from django.db import models

class MyClass(models.Model):
    name = models.CharField(max_length=64, primary_key=True, default=None)
    ip = models.GenericIPAdressField(default=None, unique=true)
    
    def __str__(self):
        return self.name
 ```
 
and `admin.py`:
```py
from django.contrib import admin
from django.contrib.auth.admin import GroupAdmin as BaseGroupAdmin
from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
from django.contrib.auth.models import Group, User
from app.models import MyClass
from unfold.admin import ModelAdmin
from unfold.forms import AdminPasswordChangeForm, UserChangeForm, UserCreationForm

admin.site.unregister(User)
admin.site.unregister(Group)


@admin.register(Server)
class MyClassAdmin(ModelAdmin):
    pass
```

When going to admin panel, we ran into exception `'str' object has no attribute 'opts'" in the unfold/templatetags/unfold.py, line 900, in tabs_primary_active`

lukasvinclav:
This is not what I meant. Please create a new issue where you will fill all questions (this is important). I can tell you right now, that your repro example is not enough because I don't see there anything special what could cause this problem so please invest little bit more time into it. I'm not saying you are not right and the error is not there, I just don't see it yet.

Issue #1844, opened 2026-02-05
Fieldsets tab are active when it has classes tab and gettext_lazy


baskarupro:
**What version of Unfold are you using?**
0.78.1

**What version of Django are you using?**
4.2.11 

**What browser are you using?**
google chrome, safari

**Did you checked changelog/commit history, if the bug is not already fixed?**

Yes

**Did you searched other issues, if the bug is not already fixed?**
Yes

**Did you checked documentation?**

Yes,

**Are you able to replicate the bug in the demo site?**

No

**Repository with reproduced bug**

Create minimal repository where it is possible to reproduce an error.

**Describe your issue**

```
from django.utils.translation import gettext_lazy as _

class MyAdmin:
        fieldsets = (
                (
                    None,
                   .....
                ),
                (
                    _("Engine"),
                  ....
                ),
                (
                    _("Transmission & Wheels"),
                    ....
                ),
```

so whe i'm trying to make it translatable, all tab are active by default.

<img width="1021" height="248" alt="Image" src="https://github.com/user-attachments/assets/1b5880cf-bb54-493c-b62b-ea310cb0bf95" />

But if im just using strings for tab names it works as expected.

<img width="1205" height="329" alt="Image" src="https://github.com/user-attachments/assets/17229c72-46e5-44bd-9bea-6cfef551018a" />

it seems like something wrong with  slugify in fieldsets_tabs.html
 <a x-on:click="activeFieldsetTab = '{{ fieldset.name|slugify_unicode }}'" x-bind:class="activeFieldsetTab == '{{ fieldset.name|slugify_unicode }}' && 'active'">

baskarupro:
<img width="973" height="156" alt="Image" src="https://github.com/user-attachments/assets/d7d5355c-d5cc-4a91-b403-735589dbddcb" />

so if i set it in english works, but if setting in another language, then slugify return ''.

PR #1880, opened 2026-02-19
fix: unicode slugify fieldsets tab attributes

madatbay:
Added test case for #1844


madatbay:
Default django slugify converts values to ASCII and this led to empty string chars in fieldsets tab attributes. As a result all tabs shown as `active`. Added custom filter which alllows unicode to solve the issue

lukasvinclav:
Thanks!

PR #1972, opened 2026-04-13
Add `filter_title` block to `admin/filter.html`

ollehu:
Add `filter_title` block to `admin/filter.html` (allows overriding the "By ..." prefix without copying full template)

No behavior change by default.


lukasvinclav:
This is already implemented here (not merged yet):

https://github.com/unfoldadmin/django-unfold/pull/1935

ollehu:
> This is already implemented here (not merged yet):
> 
> #1935

Look at that, nice :+1:  Closing this as a duplicate then.

PR #1973, opened 2026-04-13
Fix possible unbound variable issue in skeleton

tomdtp:
Fixes self-referential template issues that showed up with `django-fastdev` enabled.

Also the regular changelist now follows Django’s own `action_form` / `cl.show_admin_actions` contract again, while dataset views get an explicit `has_action_form` flag so they only show bulk-action UI when actions are actually available.


lukasvinclav:
Thanks for PR. Can you provide a description how can I replicate the issue so I can test it on local machine? I'm particularly interested in testing `unfold_admin_actions`.

btw you forget to fill the template questions ;-) Did you used AI to create this code?

tomdtp:
I separated the 2 issues, so this is now just a fix for the unbound self referencing variables.

To reproduce this you can add `django-fastdev` to the project and then add `FASTDEV_STRICT_TEMPLATE_CHECKING = True` to the `settings.py`. You will then directly see the unbound variable error for `is_fullwidth`. (if you are on main, from this branch it is fixed) if you are unauthenticated and go to the login page of the django admin.

sorry, I did not see any template questions. I just checked the other open PRs and also couldn't see any templates. so if you have questions please ask away :) I did use AI to review the code change (because I wanted it to doublecheck the default values)

the extracted issue is a contract issue, I will issue a separate pr for it

tomdtp:
any chance this can get merged? it is very disruptive in my dev workflow if the login page with `django-fastdev` keeps showing errors because of unbound template variables

lukasvinclav:
@tomdtp hey Tom, as a team of one, it is pretty hard to go through everything and do a proper testing. So right now, there is no estimate when I will have time to think what I will do with this PR.

tomdtp:
@lukasvinclav ok, I understand. thanks for the great work on unfold it is such an improvement over the default django admin!

lukasvinclav:
@tomdtp I hope I didn't let you down. The "problem" here is that I have no experience with the plugin which you are using so I have to get familiar with it. Anyway, if you find something else don't feel discouraged and just post the issue here.

tomdtp:
@lukasvinclav no, no worries, it is not a critical issue in the sense that something breaks on production. `django-fastdev` just helps you during development by showing errors when you have unbound variables in a template. Django silently just shows nothing, which can lead to frustrating errors. Since I have `django-fastdev` enabled during development I cannot open unfold without logging in via our main login page first, because the unfold login page has these unbound variables.

I am not affiliated with `django-fastdev` in any way, I just think it is a very helpful lib to spot possible problems / errors early on.

lukasvinclav:
@tomdtp sorry that it took a while. btw cool change!

tomdtp:
@lukasvinclav no worries, and thanks for reviewing and merging!
and most of all thank you again for maintaining this absolute beast of an improvement to django 🚀

PR #1982, opened 2026-04-17
Add 'f' shortcut to open changelist filters

Quentin-Jadoul:
**Description**
This PR adds a keyboard shortcut on changelist pages to open the filters panel.

**Changes**
- Added a `changeList()` Alpine helper in `src/unfold/static/unfold/js/app.js`
- Added an `f` shortcut to open filters
  - works case-insensitively
  - ignored when typing in inputs, textareas, or contenteditable elements
  - only active when filters are available (`cl.has_filters`)
- Wired the changelist root container to `changeList(...)` in `src/unfold/templates/admin/change_list.html`
- Added an `f` shortcut hint next to the Filters button
- Kept existing `Esc` behavior for closing filters

**Why**
- Makes filters quicker to open on changelist pages
- Keeps shortcut behavior aligned with existing shortcuts like `t` for search

**Notes**
- Only applies to changelist filters
- No changes outside changelist pages

image

### DIFF
--- a/src/unfold/static/unfold/js/app.js
+++ b/src/unfold/static/unfold/js/app.js
@@ -181,6 +181,28 @@ const moveRecords = (e) => {
 };
 
 /*************************************************************
+ * Change list filters
+ *************************************************************/
+function changeList(hasFilters = false) {
+  return {
+    hasFilters,
+    filterOpen: false,
+    applyShortcut(event) {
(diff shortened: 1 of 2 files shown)


lukasvinclav:
Implemented in another PR but thank you very much for good idea.

PR #1983, opened 2026-04-19
fix: django 6.1 template tag parameter

anze3db:
Django 6.1 introduced a backwards incompatible change to InclusionAdminNode that makes django-unfold crash, see this trac ticket for more info: https://code.djangoproject.com/ticket/37048

I don't have enough context to figure out if we could remove the usage of InclusionAdminNode so this change just makes it work on both versions.

I didn't add a test because existing tests will catch this when run on 6.1.

### DIFF
--- a/src/unfold/templatetags/unfold_list.py
+++ b/src/unfold/templatetags/unfold_list.py
@@ -2,6 +2,7 @@ import datetime
 from collections.abc import Generator
 from typing import Any
 
+from django import VERSION as DJANGO_VERSION
 from django.contrib.admin.options import IS_POPUP_VAR
 from django.contrib.admin.templatetags.admin_list import (
     ResultList,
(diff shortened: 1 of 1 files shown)


lukasvinclav:
Thanks for the PR. I will test it once there will be rc or beta for 6.1.

PS: How did you found this issue? Are you using Unfold with development versions of Django? I'm asking to understand your use case.

anze3db:
@lukasvinclav yes, I was testing the new Django 6.1 features on a codebase that also uses django-unfold and had to fix this issue to get it to work. 

I initially wanted to fix the backwards incompatible change on Django side but since `InclusionAdminNode` isn't public API the ticket to do that was closed as won't fix.

The first 6.1 alpha/feature freeze is going to come out on May 20 so plenty of time still before this will actually impact anyone, but I wanted to give you all a heads up. Perfectly fine to wait until beta/rc before merging but I think it's unlikely the backwards incompatible change will get reverted so this PR will be necessary.

lukasvinclav:
Thanks!

PR #2009, opened 2026-04-30
fix(sidebar): hide badge wrapper when callback returns empty string

### LINKED ISSUE #2008: Empty red badge renders in sidebar when badge callback returns an empty string
Did you used AI to write this issue?

No

What version of Unfold are you using?

0.90.0

What version of Django are you using?

6.0.4

What browser are you using?

Chrome (latest), but it's not browser-specific — it happens on every browser because it's a template issue.

Did you checked changelog/commit history, if the bug is not already fixed?

Yes

Did you searched other issues, if the bug is not already fixed?

Yes

Did you checked documentation?

Yes

Are you able to replicate the bug in the demo site?

I can't reproduce it on the demo because the demo's sidebar callbacks always return a value. But it shows up on any project the moment a badge callback returns "".

Describe your issue

When I use a sidebar item with a badge callback, the callback returns "" when there's nothing to show (e.g. zero unread messages). I expected the badge to disappear in that case, but instead the red box still renders, just with no text inside. So the sidebar shows an empty red rectangle.

I traced it and it looks like the issue is in unfold/templates/unfold/helpers/app_list_badge.html. The outer guard is:

```
{% if item.badge %}
    <span class="...bg-red-500...">
        {% if item.badge_callback %}{{ item.badge_callback }}{% else %}{{ item.badge }}{% endif %}
    </span>
{% endif %}
```

The problem: in unfold/sites.py (_get_navigation_items), when badge is a string Unfold imports it and sets item["badge_callback"] = lazy(callback)(request), but it leaves item["badge"] as the original dotted-path string (e.g. "myapp.badges.unread_count"). That string is always truthy, so {% if item.badge %} is always true, and the red <span> always renders — even when the callback returns "".

So callbacks have no way to opt out of rendering. Returning "" or None doesn't help because the guard never evaluates the resolved value.

Image

mezo-dev:
Fixes [#2008 — Empty red badge renders in sidebar when badge callback returns an empty string](https://github.com/unfoldadmin/django-unfold/issues/2008)

## Problem

When a sidebar item's `badge` is configured as a dotted-path string (callback), the red badge wrapper is always rendered — even when the callback returns an empty string. This leaves an empty coloured rectangle in the sidebar.

## Root cause

In `unfold/sites.py`, when `item["badge"]` is a string, the callback is imported and stored as `item["badge_callback"] = lazy(callback)(request)`, but `item["badge"]` is left as the original dotted-path string.

In `unfold/templates/unfold/helpers/app_list_badge.html`, the wrapper guard is:

```django
{% if item.badge %}
    <span class="...">
        {% if item.badge_callback %}{{ item.badge_callback }}{% else %}{{ item.badge }}{% endif %}
    </span>
{% endif %}
```

`item.badge` here is the path string (e.g. `"myapp.badges.unread_count"`), which is always truthy — so the `<span>` always renders. When the callback resolves to `""`, the inner content is empty but the wrapper stays, producing the empty box.

## Fix

Resolve the lazy callback to a string first and only render the wrapper when the resolved text is non-empty. Behaviour for literal-string badges (no callback) is unchanged.

## Screenshots

**Before:** empty red box next to the menu item when count = 0.
- image

**After:** no badge at all when count = 0; badge with text when count > 0.
- Screenshot 2026-04-30 at 1 41 37 PM


mgrdcm:
@mezo-dev Do you think this also solves your problem?  https://github.com/unfoldadmin/django-unfold/pull/2084

Issue #1883, opened 2026-02-19
Django permissions in actions lead to project bootstrapping problems


cfra:
**What version of Unfold are you using?**

0.80.2

**What version of Django are you using?**

6.0.2

**Repository with reproduced bug**

https://github.com/cfra/unfold-reproduction/tree/reproduction/check-issue

**Describe your issue**

When there is an admin action referencing a django permission, `django-unfold` will perform a check to see if that permission exists.

This check was introduced here:

> commit f728bd6b89950b7fa64fcb1580137de9602ec260
> Author: Lukas Vinclav <lukasvinclav@gmail.com>
> Date:   Tue May 20 14:57:03 2025 +0200
> 
>     feat: allow django permissions in actions (#1230)

For an existing deployment this works fine. Therefore, although we are using these permissions for quite some time now, we never had any issues.

However, when we try to create a new deployment, this check is problematic, because the migration command will fail as it tries to access the `auth_permission` table which has not been created yet at that time.

The same problem happens during our pre-migration sanity check:

```
./manage.py check --deploy --fail-level WARNING
```

<details>
<summary>Traceback</summary>

```
  File "/home/user/github/unfold-reproduction/./manage.py", line 22, in <module>
    main()
    ~~~~^^
  File "/home/user/github/unfold-reproduction/./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/user/github/unfold-reproduction/.venv/lib/python3.14/site-packages/django/core/management/__init__.py", line 443, in execute_from_command_line
    utility.execute()
    ~~~~~~~~~~~~~~~^^
  File "/home/user/github/unfold-reproduction/.venv/lib/python3.14/site-packages/django/core/management/__init__.py", line 437, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/user/github/unfold-reproduction/.venv/lib/python3.14/site-packages/django/core/management/base.py", line 420, in run_from_argv
    self.execute(*args, **cmd_options)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2030, opened 2026-05-10
fix: prevent OperationalError during initial migration in system checks

subhranil-179:
## Summary

This PR fixes a circular dependency in `unfold/checks.py` that causes an `OperationalError: no such table: auth_permission` during the initial project bootstrap.

## The Problem:
The system check currently queries the Permission model during the migrate or runserver command. On fresh installations, this query fails because the database schema has not yet been initialized.

## The Fix:
Added a guard clause using db_connection.introspection.table_names() to verify the existence of the auth_permission table before attempting the query. This allows migrations to complete successfully. Once the table is created, subsequent checks will run as intended.

**Closes:** #1883 

## Test plan
1. **Fresh Environment:** Deleted db.sqlite3.
2. **Migration Test:** Ran `uv run manage.py migrate` without the `--skip-checks` flag.
3. **Result:** Migration completed successfully without the OperationalError.
4. **Regression Test:** Ran `uv manage.py check` after migration to ensure the permission logic still executes correctly when the table exists.

## Use of AI

No.


lukasvinclav:
where test?

subhranil-179:
> where test?

I've added a test mocking the DB introspection in tests/test_checks.py to verify the fix on a fresh install.

subhranil-179:
Hi @lukasvinclav, does it require any other efforts? If it does let me know, else can we please merge it?

PR #2031, opened 2026-05-10
fix: center and prevent repetition of user avatars

subhranil-179:
## Summary

**Improved the avatar rendering logic in the navigation bar.**

Previously, the avatar container used **bg-cover** *only* but lacked properties to handle non-square images correctly resulting in off-center profile pictures.

## Changes:
- Added bg-center to ensure the subject of the photo remains focused in the center of the avatar circle.
- Added bg-no-repeat to prevent image tiling.

## Visual Comparison
**Before:** Screenshot 2026-05-11 at 04-24-55 29 l Change user Django site admin

**After:**  Screenshot 2026-05-11 at 04-24-35 29 l Change user Django site admin

## Test plan
1. Added a profile_pic field and a custom property avatar_url to the User model (only for testing, not commited).
2. Uploaded a wide landscape image as a profile picture.
3. Verified the navbar avatar centers on the middle of the image instead of the bottom-left corner.

## Use of AI
Nope!


lukasvinclav:
Thanks.

PR #2038, opened 2026-05-14
CSP nonce support: add nonce to style/script tags if it exists

mhindery:
## Summary

This adds the CSP nonce on the included script and style tags if it exists. This allows people to use a CSP config, which is a security mechanism controlling which assets a browser will load. CSP is a fairly modern security feature which should be set.

CSP is set on the backend via Django settings or via custom middleware:
- Django has built-in support for this starting with Django v6.0. See https://docs.djangoproject.com/en/6.0/howto/csp/#nonce-config on how to use it.
- It could also be handled with a third-party library like Secure. See https://github.com/TypeError/secure/blob/main/docs/headers/content_security_policy.md#nonce--strict-dynamic-example-recommended-pattern on how to use that.

On the template side, what is necessary is to include a `nonce=X` pair on every _script_ or _style_ tag. This PR adds those nonces on these tags in the unfold templates. It checks if the `cps_nonce` variable is set in the context, which is the value used by Django's built-in CSP mechanism.

## Test plan

/

## Use of AI

Nope, all me


lukasvinclav:
@mhindery is this already incorporated in default django admin? I'm not able to find it in templates.

mhindery:
Hey @lukasvinclav , it is in the process of being added, see the ticket on: https://code.djangoproject.com/ticket/36825 (and their github PR on https://github.com/django/django/pull/21270/changes)

mhindery:
Hey @lukasvinclav would you be able to have a look at this?

lukasvinclav:
@mhindery I'm aware of this PR but I can not promise you when I can take a look at it. If you can not wait, you can reach me here https://unfoldadmin.com/consulting/

mhindery:
Why was this closed? And without any comment or explanation?

PR #2064, opened 2026-05-30
feat: dropped django 5.1

lukasvinclav:


mgrdcm:
[Makes sense!](https://www.djangoproject.com/download/#supported-versions)

PR #2065, opened 2026-05-31
`fieldset_row_classes`: Clean up handling of missing `line` in context

mgrdcm:
## Summary

Use more direct default setting in `fieldset_row_classes` method, and make the whole method work when it's the default.  See https://github.com/unfoldadmin/django-unfold/pull/2065#issuecomment-4622368261 for more details.

## Test plan

Honestly, just did local testing.

## Use of AI

None.

### DIFF
--- a/src/unfold/templatetags/unfold.py
+++ b/src/unfold/templatetags/unfold.py
@@ -395,7 +395,7 @@ def fieldset_row_classes(context: RequestContext) -> str:
     ]
 
     formset = context.get("inline_admin_formset", None)
-    line = context.get("line") or []
+    line = context.get("line", [])
 
     # Hide the field in case of ordering field for sorting
     for field in line:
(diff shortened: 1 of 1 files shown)


lukasvinclav:
What kind of error you were getting?

mgrdcm:
> What kind of error you were getting?

Because `line` defaults to `[]` at https://github.com/unfoldadmin/django-unfold/blob/c96c0ce2a683426aa72904b256080f0eb5d6c62d/src/unfold/templatetags/unfold.py#L398 when it got down to https://github.com/unfoldadmin/django-unfold/blob/c96c0ce2a683426aa72904b256080f0eb5d6c62d/src/unfold/templatetags/unfold.py#L410 it would err with:
```
AttributeError: 'list' object has no attribute 'fields'
```

And of course it would fail a few lines later when it tries to access `has_visible_field` as well.

mgrdcm:
Happy to change this PR to use `getattr` on `line` instead of `line or ` if you'd prefer.  Was going for the most straightforward PR, but always try to match the pattern of the existing code or author's preference!

lukasvinclav:
Thanks for PR. Actually PR code led me to the implementation which was which fixes bunch of addtional typing issues https://github.com/unfoldadmin/django-unfold/pull/2077

PR #2084, opened 2026-06-07
fix: Only show badge wrapper when there's a non-empty badge value.

### LINKED ISSUE #2008: Empty red badge renders in sidebar when badge callback returns an empty string

mgrdcm:
## Summary

This fixes #2008 and the code change itself is a little simpler than #2009.  See those for details.

## Test plan

Run the tests (`uv run -- pytest .`).

## Use of AI

`None`!

# Consideration

I think it would be more appropriate to hide only the callback returns a literal `None` but that seems less backwards-compatible.

### DIFF
--- a/src/unfold/templates/unfold/helpers/app_list_badge.html
+++ b/src/unfold/templates/unfold/helpers/app_list_badge.html
@@ -1,4 +1,4 @@
-{% if item.badge %}
+{% if item.badge and item.badge_callback != '' %}
     <span class="font-semibold h-[18px] leading-[18px] ml-2 px-1 relative rounded-xs text-center text-[11px] whitespace-nowrap uppercase min-w-[18px]
                  {% if item.badge_style == "solid" %}
                      text-white text-shadow-xs
--- a/tests/test_sidebar_navigation.py
+++ b/tests/test_sidebar_navigation.py
@@ -6,9 +6,6 @@ from django.test.utils import override_settings
 from django.urls import reverse_lazy
 from django.utils.translation import gettext_lazy as _
 
-from django.contrib.auth.models import User
-from django.test import Client
-
 from unfold.settings import CONFIG_DEFAULTS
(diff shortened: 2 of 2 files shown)


lukasvinclav:
Thanks for PR which gave me an idea how to fix this issue. You can find fixed code here https://github.com/unfoldadmin/django-unfold/pull/2133

PR #2087, opened 2026-06-09
chore: typing of translation strings in decorators

MarcoSteinacher:
## Summary

Currently, the `action` and `display` decorators expect a `str` for the description argument. However, the description is often provided as a translated string with `gettext_lazy()`. This leads to type warnings, like the following in the PyCharm IDE:

grafik

This PR replaces the type hint `str` with `StrOrPromise` from `django_stubs_ext` to allow `_StrPromise` for the `description` and `label` arguments of the decorators.

## Test plan

I tested it in our actual project, which uses the decorator as such:

```
from django.utils.translation import gettext_lazy as _
[...]
@action(
        description=_("Export"),
        permissions=["geno.canview_member"],
        icon="download",
    )
```
There is no warning anymore and the type hint is correct:

grafik

Also, I ran the standard test suite of the django-unfold project.

## Use of AI

I did not use any AI tools for this PR.

### DIFF
--- a/src/unfold/decorators.py
+++ b/src/unfold/decorators.py
@@ -8,6 +8,7 @@ from django.db.models import Model
 from django.db.models.expressions import BaseExpression, Combinable
 from django.http import HttpRequest, HttpResponse
 from django.shortcuts import render
+from django_stubs_ext import StrOrPromise
 
 from unfold.dataclasses import Action, ActionDialog
 from unfold.enums import ActionVariant
@@ -18,7 +19,7 @@ def action(
     function: Callable | None = None,
     *,
     permissions: Iterable[str] | None = None,
-    description: str | None = None,
+    description: StrOrPromise | None = None,
     url_path: str | None = None,
     attrs: dict[str, Any] | None = None,
     icon: str | None = None,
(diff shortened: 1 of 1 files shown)


lukasvinclav:
@MarcoSteinacher Does it require `django-stubs-ext` in dependecies?

MarcoSteinacher:
> @MarcoSteinacher Does it require `django-stubs-ext` in dependecies?

@lukasvinclav Yes, you are right. The first commit would require `django-stubs-ext` to be added to the dependencies. We can also keep it as dev-dependency only. I made another commit for that. 

It's up to you which route to go. And anyways, this PR is not crucial to me, but it would be nice to have the correct type checking in the IDE :sunglasses:

lukasvinclav:
@MarcoSteinacher I guess we need to install `django-stubs-ext` into dev dependencies in `pyproject.toml`, right?

MarcoSteinacher:
@lukasvinclav I don't think we need to add it, because it is already pulled in by `django-stubs`, which is in the dev dependencies:

```
$ uv tree --invert --package django-stubs-ext
Resolved 76 packages in 0.85ms
django-stubs-ext v6.0.3
└── django-stubs v6.0.3
    └── django-unfold v0.97.0 (group: dev)
```

It is in the current `uv.lock` file.

lukasvinclav:
@MarcoSteinacher Thanks! PR was already released.

PR #2089, opened 2026-06-10
fix: Use request.user to not collide with user context on other views

dfr-exnaton:
## Summary

When a view is loading a `user` into the context, it also affects the rendering of the user information in the navigation. The avatar already correctly uses `request.user`. 

## Test plan

Haven't tested it

## Use of AI

It helped me find out the potential issue. I manually changed the few lines though.


lukasvinclav:
Thanks!

PR #2125, opened 2026-06-29
feat: support variant in action row

ahivert:
## Summary

  - Added support for action variants on inline changelist row actions.
  > Previously, row actions rendered outside of the dropdown via `extra_options={"display_in_dropdown": False}` always used the default button variant. Inline row actions now pass the configured `ActionVariant` to the button component, so variants such as `DANGER`, `PRIMARY`, or `SUCCESS` are rendered correctly.
  
- Updated the changelist row actions documentation with an inline row action example using `variant=ActionVariant.DANGER`.

## Test plan

  - Added a test covering an inline changelist row action with `ActionVariant.DANGER`.
  - Verified the rendered changelist contains the row action label.
  - Verified the rendered output contains the expected danger button styling.
  
## Use of AI

AI was used for all changes done.

### DIFF
--- a/docs/actions/changelist-row.md
+++ b/docs/actions/changelist-row.md
@@ -42,3 +42,35 @@ class UserAdmin(ModelAdmin):
         # Write your own bussiness logic. Code below will always display an action.
         return True
 ```
+
+Row actions can also be displayed directly in the changelist row instead of inside
+the dropdown menu by setting `extra_options={"display_in_dropdown": False}`.
+Inline row actions support the same `variant` values as other action buttons.
+
+```python
+# admin.py
+
+from django.contrib.admin import register
+from django.contrib.auth.models import User
+from django.shortcuts import redirect
+from django.urls import reverse_lazy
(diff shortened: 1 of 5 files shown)


ahivert:
Hey @lukasvinclav I just rebase to ensure no conflicts. Do I need to change something to make this PR move on? 

It's literary 2 lines change (excluded test), so real quick win.

lukasvinclav:
Hey Anthony, thanks for the PR. I tested the functionality and it does not work. Color of the button does not change based on the variant because button component does not support colors.

ahivert:
> Hey Anthony, thanks for the PR. I tested the functionality and it does not work. Color of the button does not change based on the variant because button component does not support colors.



Ok fair, "danger" was already there but not the others. My bad I should have checked others variants. 

I updated the button component to add missing variants "success", "info", "warning". 

Button now accepts all variants like other components (labels and badges).

Now the example admin display success button and danger button. 

image

Can we reopen this PR? I'd like a last chance to get it merged with the fix.

PR #2136, opened 2026-07-06
feat: trap focus in filter panel when explicitly opened

mgrdcm:
## Summary

Contains focus to the filter panel when opened, allowing tabbing through focusable elements in the filter panel.  As things are, especially useful in FireFox but works in Safari too for actual form inputs.

Inspired by the recent keyboard shortcuts added and especially https://github.com/unfoldadmin/django-unfold/pull/2131

Depends on https://github.com/unfoldadmin/django-unfold/commit/f8c71585694a020305fe304bce2f2dcb03f18c10

## Test plan

It's all browser interaction, and since there aren't any Selenium tests or anything, just manually.  You can open the filter panel with either the Filters button or the fancy new `f` shortcut, then tab through.  In FireFox you can focus on links even and then just hit enter/return, in Safari only the actual form inputs will be useful now.

## Use of AI

`None`


mgrdcm:
BTW if want to make a little simpler
```html
x-data="{filterAlwaysOpen: getComputedStyle(document.getElementById('changelist-filter-close')).display == 'none'}"
```
could _maybe_ be
```html
x-data="{filterAlwaysOpen: null}"
```
because in my testing the `x-resize` handler was called even on initial page load, but I couldn't find any documentation saying the resize event will always trigger on initial page load so this is a little safer.

mgrdcm:
Replaced by https://github.com/unfoldadmin/django-unfold/pull/2145

PR #2154, opened 2026-07-17
feat: redesign changelist table

lukasvinclav:


swayll:
Previously, long text was displayed in several lines, but now it is displayed in one line with a long scroll - this is very inconvenient!

lukasvinclav:
@swayll why the exclamation mark at the end of the sentence?

swayll:
> @swayll why the exclamation mark at the end of the sentence?

Sorry, but this is an unexpected and strange innovation that was introduced suddenly and worsens the user experience, and it is not clear what thoughts led to removing the multi-line representation of data in tables?

PR #2168, opened 2026-07-20
feat: added changeform_fullwidth option to ModelAdmin

ciaransanders:
## Summary

Add new option on `src.unfold.admin.ModelAdmin` called `changeform_fullwidth`. When set to `True`, the changeform page for the corresponding `ModelAdmin` has the `container` class remove, allowing the page to be full width.

## Test plan

Visually tested, by trying all permutations of True/False for both `list_fullwidth` and `changeform_fullwidth` to verify they do not interfere with each other.

## Use of AI

Yes, AI was used to help determine a method by which we could have `list_fullwidth` and `changeform_fullwidth` not accidentally interfere with each other in the logic of the `if` statement. 

## Improvements

This implementation depends on the fact that `cl` is only available on the change list view and `adminform` is only available on the change form page. Not my favorite idea, personally... but it works, so I'm just going to create this draft pull request for the moment and I can iterate on the idea later; especially if someone has suggestions!

### DIFF
--- a/docs/configuration/modeladmin.md
+++ b/docs/configuration/modeladmin.md
@@ -50,6 +50,9 @@ class CustomAdminClass(ModelAdmin):
     # Display changelist in fullwidth
     list_fullwidth = False
 
+    # Display changeform in fullwidth
+    changeform_fullwidth = False
+
     # Set to False, to enable filter as "sidebar"
     list_filter_sheet = True
 
--- a/src/unfold/admin.py
+++ b/src/unfold/admin.py
@@ -66,6 +66,7 @@ class ModelAdmin(
     list_filter_sheet = True
     list_filter_options: dict[str, ListFilterOptionsItem] = {}
     list_fullwidth = False
+    changeform_fullwidth = False
     list_disable_select_all = False
     list_before_template = None
     list_after_template = None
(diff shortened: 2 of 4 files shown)


lukasvinclav:
I was already considering this feature. My question is why you need that? May you describe your issue with the current non-fullwidth layout?

ciaransanders:
Yeah that's a good question, because I agree that the current width is perfect for 99% of use cases. 

But in my scenario, I have a high traffic admin page which has a couple of needs for more screen space. 

First, It contains a "step" inline. This step inline is sortable, must be a tabular inline, and now has too many fields to fit within the screen width. I _could_ just switch over to using a stacked inline, but the issue is a lot of the objects on this page have a _lot_ of steps which the user wants to easily be able to view, and they by and large prefer the stacked inline for this purpose.

This admin page also has a customized template displaying images, for reference and design purposes. Having the page be full width would allow us to display the images larger; a small QOL thing.

It's probably something I could work around, and just keep this admin page the current width. But setting the page as full width would be a nice feature, at least I would find use in it 👍

lukasvinclav:
I think tabular inline with a lot of columns is good enough example for me. I will take a look at this but as always, I'm not promising anything.

PR #2171, opened 2026-07-22
fix: reuse changelist result headers

matevz-ap:
## Summary

Generate changelist result headers once and reuse them while rendering each result row.

### Problem

`items_for_result()` regenerated all headers for every result row. Header generation performs field inspection, sorting calculations, widget rendering, and three query-string constructions for every sortable column.

For a page containing 50 rows and four sortable columns, this resulted in 612 query-string constructions instead of 12.

### Changes

- Pass the already-generated headers from `result_list()` into `results()`.
- Pass those headers into `items_for_result()`.
- Preserve compatibility for callers that invoke `results()` or `items_for_result()` without supplying headers.

## Test plan

A local 50-row benchmark with four sortable columns showed:

- Before: approximately 6.99 ms
- After: approximately 4.95 ms
- Query-string constructions: 612 → 12
- Speedup: approximately 1.41×

Also passing all tests locally.

## Use of AI
ChatGPT 5.6 Sol pointed out to me while migrating from grappelli about this potential rendering issue. All the code was written by it.

### DIFF
--- a/src/unfold/templatetags/unfold_list.py
+++ b/src/unfold/templatetags/unfold_list.py
@@ -1,5 +1,5 @@
 import datetime
-from collections.abc import Generator
+from collections.abc import Generator, Sequence
 from typing import Any
 
 from django import VERSION as DJANGO_VERSION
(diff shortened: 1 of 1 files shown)


lukasvinclav:
How can I replicate this issue?

matevz-ap:
My understanding is that result headers should be called once to get the table headers. Adding a print statement and running `uv run pytest -s tests/test_tags.py::test_tags_result_list_tag --no-cov`  prints 2 times. Calling the function for each row seems reduntant.

lukasvinclav:
You are right. These headers are not necessarry anymore as I removed the functionality which was requiring them last week. Proper implementation is in this PR:

https://github.com/unfoldadmin/django-unfold/pull/2172

Thanks for pointing me at this issue.

PR #2192, opened 2026-08-03
Pass `check_test` callback to super for switch widget

Eroica:
Hi,

I noticed that Unfold's switch widget ignores any `check_test` callback. Is there a specific reason for this? `UnfoldBooleanWidget`, on the other hand, passes it to Django's parent, so I just did the same for `UnfoldBooleanSwitchWidget`. Let me know if that's not intended.

## Summary

* Unfold's `UnfoldBooleanSwitchWidget` sub-classes Django's `CheckboxInput`, but throws away any `check_test` callback by always calling super with `None`
* This makes it necessary to set the initial state of this widget using other mechanisms (e.g. `initials`, etc.)
* I simply mirrored what `UnfoldBooleanWidget` does: Pass the `check_test` to super

## Test plan

* Instantiated `UnfoldBooleanSwitchWidget` witch `check_test=...` callbacks.

## Use of AI

None.

### DIFF
--- a/src/unfold/widgets.py
+++ b/src/unfold/widgets.py
@@ -810,8 +810,6 @@ class UnfoldBooleanWidget(CheckboxInput):
     def __init__(
         self, attrs: dict[str, Any] | None = None, check_test: Callable | None = None
     ) -> None:
-        attrs = attrs or {}
-
         super().__init__(
             {
                 **(attrs or {}),
--- a/src/unfold/widgets.py
+++ b/src/unfold/widgets.py
@@ -832,7 +832,7 @@ class UnfoldBooleanSwitchWidget(CheckboxInput):
                     [*SWITCH_CLASSES, attrs.get("class", "") if attrs else ""]
                 ),
             },
-            check_test=None,
+            check_test,
         )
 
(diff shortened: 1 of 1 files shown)


lukasvinclav:
hey, why you removed `attrs = attrs or {}`?

lukasvinclav:
what do you think about the implementation below? Would you mind testing it?

https://github.com/unfoldadmin/django-unfold/pull/2193/changes

Eroica:
> hey, why you removed `attrs = attrs or {}`?

It's in the first commit, the code looked like this:

```python
        attrs = attrs or {}

        super().__init__(
            {
                **(attrs or {}),
```

I just thought that obtaining a default for `attrs` twice is redundant. FWIW, `UnfoldBooleanSwitchWidget` (just below it) also doesn't use an extra assignment (that is, before your new PR).

> what do you think about the implementation below? Would you mind testing it?

I took a look, it's good! Choose whatever code style you prefer.

lukasvinclav:
@Eroica I merged my PR because it contains a test. btw thanks for report!

Issue #2120, opened 2026-06-25
Relationship item overflow in other fields in object view


lays147:
**Did you used AI to write this issue?**
Yes, used to aggregate context

**What version of Unfold are you using?**

0.99

**What version of Django are you using?**

For example: 6.0.6

**What browser are you using?**

Firefox 152.0.1

**Did you checked changelog/commit history, if the bug is not already fixed?**

Yes

**Did you searched other issues, if the bug is not already fixed?**

Yes

**Did you checked documentation?**

Yes

**Are you able to replicate the bug in the demo site?**

No

**Repository with reproduced bug**

Create minimal repository where it is possible to reproduce an error.

**Describe your issue**

When accessing the object that has a foreign key, the three dots overflow above the next column in the same row:

<img width="1293" height="264" alt="Image" src="https://github.com/user-attachments/assets/8d4676ca-f243-4aa3-815b-a80ae3ad861a" />

Context from Claude:

When an admin field rendered with `autocomplete_fields` (or any field using
the `related_widget_wrapper.html` related-object widget) shares a fieldset
row with other fields (e.g. a 3-column row via `fields = [("customer",
"status", "is_paid")]`), the widget's "more options" button (the `⋮` /
`more_vert` icon, fixed at `w-[38px] h-[38px]`) overflows out of its grid
column and visually overlaps the next field in the row instead of staying
within its own column.

## Steps to reproduce
1. Define a `ModelAdmin` fieldset with 3+ fields on one line, where the
   first field is an FK rendered via `autocomplete_fields`:
   ```python
   fieldsets = (
       ("Section", {"fields": [("customer", "status", "is_paid")]}),
   )
   autocomplete_fields = ["customer"]
   ```
2. Open the change form for that model in the admin.
3. Observe the row: the autocomplete input for `customer` renders correctly,
   but the `⋮` button next to it overlaps/floats above the `status` field
   that follows in the same row, instead of sitting flush beside the input
   within the `customer` column.

## Root cause analysis
- `unfold/templatetags/unfold.py::fieldset_row_classes` assigns
  `grid lg:grid-cols-{N}` to a fieldset row when it has N > 1 fields,
  splitting the row into N equal-width columns.
- `unfold/templates/unfold/widgets/related_widget_wrapper.html` renders the
  related field as `flex flex-row` containing the actual input/select
  (which should grow/shrink) plus a sibling `div` fixed at
  `w-[38px] h-[38px]` for the "more options" trigger.
[...]

PR #2244, opened 2026-08-24
fix!: disable changeform line fields

lukasvinclav:


MrSmitix:
Hello. Could you please explain why this feature is being removed? I find FieldSets a very useful tool for creating more compact and user-friendly admin panel pages. If the issue is described in #2120, perhaps we can find a less radical solution?

lukasvinclav:
@MrSmitix what is a less radical solution?

MrSmitix:
The solution where you can describe 2 and more column in one line

lukasvinclav:
@MrSmitix how do you know that the original implementation is not radical?

MrSmitix:
@lukasvinclav Are you referring to the original unfold implementation or django.contrib.admin? I'm simply saying that fieldsets are an extremely useful tool for field composition, and I don't think it's right to simply remove support for them. Especially given that they're preserved in the original Django, and you claim to "preserve familiar models, permissions, and conventions." I'd like to know if we can consider a different solution to the problem described in 2120 than removing support for multiple field columns in a single row. Or perhaps I could understand your motivations and reasoning behind this decision, if I'm missing anything.

lukasvinclav:
@MrSmitix what do you think why I removed it?

MrSmitix:
@lukasvinclav I don't know

lukasvinclav:
@MrSmitix at the moment, I don't have enough resources to maintain this functionality. There is still an option to override the template in your project and do the customization

fng3r:
It's quite sad that support for fieldsets as a way to organize fields layout was dropped. I understand why you might want to do this but this was a great capability to utilize horizontal space to create more compact and ergonomic ui.
Btw, fieldsets worked fine with "uncompressed fields" which support was dropped earlier as well. Maybe you would be willing to reconsider this in the future

lukasvinclav:
@fng3r Dmitry, thank you very much for the PR yesterday. It was a good catch.

Now, regarding your comment, what you are suggesting? I'm open to hear a long term plan which will work with the current resources I have.

Issue #2230, opened 2026-08-20
Collapsible stacked inline renders x-show="open" without any x-data (Alpine "Illegal invocation" warnings)


miroslavudan:
Version: django-unfold 0.104.1 (also present on `main` at the time of writing), Django 6.0, Python 3.13.

### What happens

A **non-nested** collapsible `StackedInline` (`classes = ["collapse"]`) logs two Alpine warnings per inline on every change page:

```
Alpine Expression Error: Illegal invocation
Expression: "open"
```

### Why

In `src/unfold/templates/unfold/helpers/edit_inline/stacked_list.html` the two sides of the same feature are gated differently:

- line 3 declares `x-data="{ open: ... }"` only `{% if inline_admin_formset.is_collapsible and nested_inline %}`
- lines 10 and 11 render `x-show="open"` `{% if inline_admin_formset.is_collapsible %}` — without the `nested_inline` part

So for a non-nested collapsible inline the two elements evaluate `open` with no Alpine scope defining it. Alpine falls back to `window.open`, which throws `Illegal invocation` when read as a getter, and the expression is truthy anyway.

`src/unfold/templates/unfold/helpers/edit_inline/tabular_table.html` gates both on `is_collapsible and nested_inline`, which is why tabular inlines are quiet.

### Impact

Cosmetic. Collapsing itself is unaffected because non-nested inlines collapse through the native `<details>`/`<summary>` wrapper in `admin/edit_inline/stacked.html`; only the console fills up (2 warnings per collapsible stacked inline per page load).

### Steps to reproduce

1. Register a `StackedInline` with `classes = ["collapse"]` on any `ModelAdmin` (not nested).
2. Open the change page and look at the browser console.

### Suggested fix

Add `and nested_inline` to the two `x-show="open"` conditions on lines 10 and 11 of `stacked_list.html`, so they match the `x-data` on line 3 and the equivalent conditions in `tabular_table.html`.

lukasvinclav:
Car Miro, just use `collapsible = True` on `StackedInline`:

https://unfoldadmin.com/docs/inlines/options/

PR #2265, opened 2026-08-31
feat: unlimited nesting depth for nested inlines

Brktrlw:
# PR: feat: unlimited nesting depth for nested inlines

> Target: `unfoldadmin/django-unfold` · branch `feat/infinite-nested-inlines`
> Everything below the line is the PR body, paste it as-is.

---

## Summary

Nested inlines were limited to a single level, and could not be nested through a many-to-many relation. Both limits are lifted here, along with the issues that only become visible once a hierarchy gets deeper.

The cause of the depth limit was in one place. `_build_nested_formsets()` walked the forms of the top level formsets and built one level of nested formsets below them, but never descended into the formsets it had just built, so a third level of `inlines` was declared and silently ignored. It is now recursive, together with `nested_all_valid()`.

The rest of the stack already worked at any depth and is left as it is: `save_formset()` already recursed, the templates include each other through `form.nested_formsets`, and the `${formsetName}-__prefix__` replacement in `app.js` is written so that placeholders at the levels below stay untouched and are resolved when a record is added there. This keeps the change close to the existing design.

```python
class ProjectAdmin(ModelAdmin):
    inlines = [TaskInline]

class TaskInline(TabularInline):
    inlines = [SubTaskInline]

class SubTaskInline(TabularInline):
    inlines = [SubTaskItemInline]   # rendered now
```

### Depth

An acyclic chain of inlines ends where the declarations end, so it needs no configuration. A chain which contains itself does not: every level also renders the template form used for adding records, so there is always one more level to render. `nested_inlines_max_depth` defines how deep such a chain goes, and a system check reports the chains that need it.

```python
class SubcategoryInline(TabularInline):
    model = Category
    fk_name = "parent"

SubcategoryInline.inlines = [SubcategoryInline]

class CategoryAdmin(ModelAdmin):
    inlines = [SubcategoryInline]
    nested_inlines_max_depth = 4
```

Reporting this through a system check (`unfold.E001`) rather than only at request time follows what Django does for the rest of the admin configuration, and it means the mistake surfaces when the application starts. The recursion keeps a guard for the case where the check has been silenced, or where the chain only appears through `get_inlines()`.

### Nesting through many-to-many

An inline rendering the intermediary model of a many-to-many relation can now nest inlines of the model on the other side of it, which previously raised `ValueError: 'app.TagNote' has no ForeignKey to 'app.User_tags'`.

```python
class TagNoteInline(TabularInline):
    model = TagNote          # foreign key to Tag

class UserTagInline(StackedInline):
    model = User.tags.through
    inlines = [TagNoteInline]
```

Traversing a relation of the parent model is done only for an intermediary model (`_meta.auto_created`). Anywhere else, an inline without a relation to its parent stays the configuration error it always was, rather than being attached to some other model which happens to be reachable.

The object the nested records belong to is resolved again while saving, after the form carrying the relation has been saved. That is what lets a record be added below a through row created by the same request, and it means the records follow the object the submitted data names rather than the one the relation pointed at while the page was rendered.

When several relations would fit, `nested_parent_field` names the one to use. The ambiguity is reported by a system check (`unfold.E002`) instead of being guessed.

### Fixes along the way

- **File fields in nested inlines were dropped on save** (#1960). Django computes `has_file_field` from `inline_admin_formsets` only, so a file field that exists solely in a nested inline never made the change form multipart and the upload never reached the server. `render_change_form()` now takes the nested formsets into account, including the ones below a template form, which are the only ones a formset has when `extra` is zero and nothing is saved yet.
- **A top level `StackedInline` with `classes = ["collapse"]` was permanently hidden** (#2230). `stacked_list.html` emitted `x-show="open"` while the matching `x-data` was rendered only for a nested inline, so Alpine evaluated an undefined value as falsy. The binding is now rendered under the same condition as the value it reads.
- **Nested formsets below a template form were bound to the submitted data.** They reported `ManagementForm data is missing or has been tampered with` for prefixes still containing `__prefix__`, once per level. They are client side templates and are no longer bound.
- **`nested_all_valid()` stopped at the first invalid formset**, which hid the errors of the ones after it. It now validates all of them, as `django.contrib.admin.options.all_valid()` does.
- **The media of the nested formsets was accumulated on the `ModelAdmin` instance**, which is created once and shared by every request being served, so it both grew without bound and could be replaced by a concurrent request. It is collected on the request and added to the change form context instead. `ModelAdmin.nested_formset_media` is gone; nothing else read it.
- `app.js`: the reindex after a delete scopes to the row's own `.form-nested` element and rebinds the handlers it recreates, including the ones of the nested add buttons; showing the add button again after a delete is scoped to the formset's own footer, since the nested formsets have their own and come first in document order; `applyLastClass()` no longer assumes a next sibling exists, which it does not when adding is not permitted; and nested **stacked** rows now receive the `formset:added` event, which only tabular tables were getting, so widgets in nested stacked rows were left uninitialized.

### A note on size

Every form renders the whole hierarchy below it, including the forms `extra` adds, so the rendered HTML multiplies with each level and `extra` is the largest factor in that product. On the four level chain added to the example project, the change form went from 2016 KB to 274 KB by setting `extra = 0` on the inlines and adding the `collapse` class. The documentation says this, and the example project uses both.

Two costs are left as they are and are worth naming. `ModelAdmin.get_formset_kwargs()` copies `request.POST` for every formset it builds, which the recursion now calls once per form per level, and `inline.get_formset()` builds a formset and a form class for each of them. Both are per request work that could be shared, and both are better addressed on their own than inside this change.

This is an alternative to #1943, which wraps the external `django-nested-admin` package as a contrib module. This one keeps the existing `inlines = [...]` API, adds no dependency, and the runtime change is contained in `NestedInlinesModelAdminMixin`.

## Test plan

The example project gained a four level chain (`Invoice → InvoiceItem → InvoiceItemPart → InvoiceItemPartNote`), a self referencing `CategoryTree` admin, and two models behind the `User`/`Tag` relation. `tests/test_inlines.py` and `tests/test_checks.py` cover:

- rendering, creating, updating and deleting at the third and fourth level
- "You can not create nested object without parent" raised at a deeper level
- `nested_inlines_max_depth` bounding a self referencing chain, and the `ImproperlyConfigured` guard behind it
- both system checks, reporting and accepting a self referencing chain and an ambiguous relation
- nesting through a many-to-many intermediary model: the detected relation, `nested_parent_field`, a record added below a through row created by the same request, and the records following a relation changed by the same request
- an inline with no relation to its parent staying a configuration error
- `enctype="multipart/form-data"` when the only file field sits in a nested inline of a formset with no rows, and a file uploaded there reaching the model
- a nested inline with the `collapse` class rendering closed, a nested formset holding errors rendering expanded, and a top level collapsible `StackedInline` not binding to an undefined value
- the media of the nested inlines reaching the page without being kept on the `ModelAdmin`

`uv run pytest` passes (460 passed, 1 skipped) with coverage at 95.66%. `tox` passes on Django 5.2, 6.0, 6.1 and main. `pre-commit run` passes on every touched file.

Verified on the example project that the change form renders four levels, that the file field makes the form multipart, and that the collapsible levels render closed.

## Use of AI

I made the decisions on the approach: making the existing mixin recursive rather than pulling in django-nested-admin, requiring an explicit depth only for self-referencing chains, reporting misconfiguration through system checks, and keeping M2M support in scope. The tests were written with Claude Code based on the behaviour and cases I defined. I also had the implementation reviewed by AI for potential issues and edge cases, then reviewed the changes myself and verified the behaviour in the browser.


lukasvinclav:
Thanks for PR. Unfortunately at the moment, unlimited nested inlines are not in the roadmap.

PR #2286, opened 2026-09-13
fix: nested stacked inline deletion

### LINKED ISSUE #2285: Deleting nested inline also triggers deletion overlay on parent inline when using StackedInline
**Did you use AI to write this issue?**

No

0.105.0

6.0.3

Chrome 152.0.7977.75

**Did you check the changelog/commit history, if the bug is not already fixed?**

Yes

**Did you search existing issues to see if the bug is already fixed?**

Yes

**Did you check the documentation?**

Yes

No. The only example of nested inlines I found on demo site is **Customer** -> **Orders** -> **Order items** but it's built with `TabularInline`

**Describe your issue**

When using nested inlines with `StackedInline`, clicking "Delete" button on child inline triggers deletion overlay for both child and parent inline. Furthemore, since parent overlay has greater z-index than child one (z-50 vs z-40) it can even be closed 'cause we can't uncheck child delete checkbox which is covered by parent overlay.

Root cause is too broad selector `has-[.delete:checked]:[&>.form-delete]:block` on [`form-group`](https://github.com/unfoldadmin/django-unfold/blob/main/src/unfold/templates/unfold/helpers/edit_inline/stacked_list.html#L14) which would check for ANY delete checkbox on arbitrary depth in the tree and nested inlines are actually nested inside parent one in markup. This is why it would trigger both child and parent overlays to show. `TabularInline` is not effected since nested inlines placed in sibling `<tr>` and not nested inside parent inline in markup

https://github.com/user-attachments/assets/46eff7d7-ce6e-41bc-8a90-4a7265f29c4f

fng3r:
## Summary

Fix https://github.com/unfoldadmin/django-unfold/issues/2285

Narrow down `has-[.delete:checked]:[&>.form-delete]:block` selector to `has-[>_.form-row_.delete:checked]:[&>.form-delete]:block` to only check  `.delete` checkbox inside direct child `.form-row` instead of going down the whole subtree

## Test plan

Tested all combinations of stacked/tabluar inlines nested into each other
Tailwind playground [demo](https://play.tailwindcss.com/ONSfHOkNR3) that reconstructs main parts of stacked inlines markup and showcases current behaviour and proposed fix

## Use of AI

Used to generate tailwind demo


fng3r:
@lukasvinclav Didn't find how you build tailwind in the project. Are you doing it manually?

PR #2295, opened 2026-09-22
fix: sidebar border in rtl

n-shadloo:
## Summary

In a right-to-left language the sidebar sits on the right, but `#nav-sidebar` draws its separator with `border-r`, so the line lands on the outer edge of the window and nothing separates the sidebar from the content. This changes that one class to `border-e`.

`border-e` is `border-inline-end`: the right side in LTR and the left side in RTL, so the LTR output does not change. The only change in the compiled CSS is one added rule, `.border-e{border-inline-end-style:var(--tw-border-style);border-inline-end-width:1px}`, 85 bytes; nothing is removed, because `border-r` stays in the `@source inline` safelist.

Persian, desktop:

| before | after |
|---|---|
|  |  |

Persian, mobile menu:

| before | after |
|---|---|
|  |  |

## Test plan

I ran the `tests/server/example` project twice against one database, once from `main` at 91b375a and once from this branch, so the two servers differ only by this change.

LTR, `en`: 160 screenshots per server in Chromium 153.0.8010.12 and WebKit 26.6 (Playwright 1.63.0), at 1440x900, 1024x768 and 390x844, light and dark, over 15 page and sidebar states, 14 of which apply at 1440 and 13 at the smaller viewports, where the sidebar starts closed: dashboard, app index, changelist with filters and pagination, changelist with rows selected, add form, change form with inlines and tabs, object history, delete confirmation, login, and the sidebar expanded, collapsed, opened from the header, with a collapsible group open, with the all applications overlay open, and with the command search in use. Compared pixel by pixel against the same matrix from `main`: 0 changed pixels.

RTL: the same matrix in `fa` and `ar`. I measured the separator in the screenshots rather than judging it by eye. Before, the border pixel is the last column of the window: x=1439 at 1440, x=1023 at 1024, x=389 at 390. After, it is at the sidebar's inline start, x=1152, 736 and 102, and the window edge is clear. Checked at 1440 in light and dark, at 1024 and at 390, in both languages and both engines.

I also compared every element's computed box between a left-to-right and a right-to-left render of the same page, keyed by DOM path. `main` fails 81380 of those comparisons, this branch 81316. The 64 that disappear are all `#nav-sidebar`'s `border-left-width` and `border-right-width`, in every run and state where the sidebar is visible. Nothing new appears, and no other failure changes.

`uv run --frozen -- pytest .`: 448 passed, 1 skipped, coverage 95.56%. `pre-commit run --all-files`: clean, no files changed. The compiled CSS was rebuilt with the standalone Tailwind 4.3.3 binary, matching the version in the committed file's header.

## Use of AI

Yes. AI was used for the code, the test tooling and this description, and I reviewed all of it myself before opening this PR.


lukasvinclav:
@n-shadloo thanks for the PR. Would you mind to remove compiled css from the PR? I will compile it on my own with your changes when I will test.

n-shadloo:
Done, removed it.

lukasvinclav:
Awesome. Thanks for contribution.

